PR #4, opened 2018-07-30
Added conical and confined axis constraints, relevent Json files, and…

reidchristopher:
… tests for the new constraints.

### DIFF
--- a/trajopt/include/trajopt/file_write_callback.hpp
+++ b/trajopt/include/trajopt/file_write_callback.hpp
@@ -1,0 +1,26 @@
+#pragma once
+#include <trajopt/common.hpp>
+#include <trajopt/problem_description.hpp>
+#include <trajopt_utils/eigen_conversions.hpp>
+#include <tesseract_core/basic_kin.h>
+#include <tesseract_core/basic_env.h>
+#include <trajopt/plot_callback.hpp>
+#include <functional>
+#include <set>
+#include <fstream>
+#include <string>
+#include <stdio.h>
+#include <Eigen/Core>
+
+typedef std::shared_ptr<std::ofstream> ofstreamPtr;
+
+using namespace Eigen;
+using namespace util;
+using namespace std;
+namespace trajopt
+{
+
+  Optimizer::Callback WriteCallback(string file_name,TrajOptProbPtr prob, vector<string> joint_names,
+                                    vector<Affine3d> pose_inverses, Affine3d tcp = Affine3d::Identity(), bool angle_axis = true, bool degrees = true);
+
+}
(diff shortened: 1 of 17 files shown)


trajopt/include/trajopt/file_write_callback.hpp
```diff
@@ -0,0 +1,26 @@
...
+#include <trajopt_utils/eigen_conversions.hpp>
+#include <tesseract_core/basic_kin.h>
+#include <tesseract_core/basic_env.h>
+#include <trajopt/plot_callback.hpp>
+#include <functional>
+#include <set>
+#include <fstream>
+#include <string>
+#include <stdio.h>
+#include <Eigen/Core>
```
Jmeyer1292: Don't put includes in a public header that aren't needed _in the header_. Move as much to the source file as possible.

trajopt/include/trajopt/file_write_callback.hpp
```diff
@@ -0,0 +1,26 @@
...
+
+typedef std::shared_ptr<std::ofstream> ofstreamPtr;
+
+using namespace Eigen;
+using namespace util;
+using namespace std;
```
Jmeyer1292: Is this meant to be a publicly accessed header? If so, do not put using directives in it. You're effectively also making the decision to use the same namespaces on behalf of all of your clients. Put it in the source file.

trajopt/include/trajopt/file_write_callback.hpp
```diff
@@ -0,0 +1,26 @@
...
+namespace trajopt
+{
+
+  Optimizer::Callback WriteCallback(string file_name,TrajOptProbPtr prob, vector<string> joint_names,
```
Jmeyer1292: Is the lack of references in this function signature intentional? Why not pass joint names as a `const std::vector<std::string>&`?

Jmeyer1292: I'm sometimes lazy about this (and I'm not sure it matters for 64bit machines), but Eigen types require certain memory alignment for their SIMD acceleration to function correctly.

So when writing `std::vector<Eigen::Affine3d>`, you should probably write instead: `std::vector<Eigen::Affine3d, Eigen::aligned_allocator<Eigen::Affine3d>>` instead. There's a package in ROS called  `eigen_stl_containers` that provides some aliases for you. It's probably better to define your own using templated aliases.

trajopt/include/trajopt/file_write_callback.hpp
reidchristopher: Definitely needs to be changed. Didn't understand the extent of what this did back then

trajopt/include/trajopt/file_write_callback.hpp
reidchristopher: I cannot remember if the pass by value was purposeful. Can you explain to me what passing Eigen's allocator does and why it is preferable?

PR #5, opened 2018-08-09
Changed scaling from coefficients in CostFromErrFunc to be linear for all penalty types.

reidchristopher:
It was previously quadratic for the SQUARED penalty type. This meant that if the coefficient values of one cost were x times larger than another, they were weighted x^2 times as strongly. This is not intuitive for the user, and inconsistent with how other cost functions are set up (such as those is trajectory_costs).


Levi-Armstrong:
It was changed

From: <a href="https://www.codecogs.com/eqnedit.php?latex=(y(a&plus;bx_{1}&plus;cx_{2}))^{2}" target="_blank"></a>

To: <a href="https://www.codecogs.com/eqnedit.php?latex=y(a&plus;bx_{1}&plus;cx_{2})^{2}" target="_blank"></a>

PR #7, opened 2018-08-10
Added file writing and graphing utilities.

reidchristopher:
This provides a callback to write the iterations of trajectories to files and a python script to compare two final trajectories via graphs.


Levi-Armstrong:
Run clang formating

mpowelson:
@Levi-Armstrong Can we merge this? It needs the affines changed to  isometry, but other than that it seems to work fine. I figured I can fix that and add an example usage in a separate PR.

Levi-Armstrong:
This was pulled into PR #20. Thanks for the contribution @reidchristopher.

PR #11, opened 2018-08-30
Use isometry

Levi-Armstrong:


trajopt_tools/package.xml
```diff
@@ -12,7 +12,7 @@
   <build_export_depend>roscpp</build_export_depend>
   <exec_depend>roscpp</exec_depend>
   <depend>trajopt_ext</depend>
-  <depend>tesseract_ext</depend>
+  <depend>bullet3_ros</depend>
```
Jmeyer1292: Is this a new package or a rosdep dependency?

Levi-Armstrong: So tesseract_ext was a package only including bullet3 but I had issues when integrating fcl so  moving into a ros package making tesseract_ext a meta package in preparation for a upcoming PR integrating FCL.

Jmeyer1292: Ready to merge?

PR #29, opened 2018-10-29
Refactor naming of costs/constraints

mpowelson:
Obviously this is a pretty invasive PR that would break people's code, so see what you think. I was planning on doing this when when we merged time param, but since that is still in progress I figured it was better to do it sooner before any more projects get kicked off using this. There are two parts to this.
1)  I renamed the costs and constraints to be consistent. I added Cart to clarify which terms are Cartesian, and I renamed everything to "term" instead of cost/constraint.

Edited to match what got merged
- **What we currently have | My proposed changes**
- PoseCostInfo | DynamicCartPosTermInfo
- StaticPoseCostInfo | CartPosTermInfo
- CartVelCntInfo | CartVelTermInfo
- JointPosCostInfo | JointPosTermInfo
- JointVelCostInfo | JointVelTermInfo
- JointAccCostInfo | JointAccTermInfo
- JointJerkCostInfo | JointJerkTermInfo
- CollisionCostInfo | CollisionTermInfo
- JointConstraintInfo |  (went away)
- JointVelConstraintInfo |  (went away)

2) I did my best to add Doxygen comments to the costs/constraints. Hopefully along with the new naming, this will help make it more obvious what everything does.

Still TODO: 
Add constraints to Joint position, velocity, acceleration, and jerk and remove JointVelConstraintInfo and JointConstraintInfo.


mpowelson:
Just added constraints to joint space terms. One other thing we should consider that this does not address is whether or not the json terms should be renamed. They are pretty good, but cart_vel/pose/static_pose should probably be made more consistent. I believe these are set in problem_description line 48-56.

trajopt/include/trajopt/kinematic_terms.hpp
```diff
@@ -13,14 +13,18 @@ namespace trajopt
 using namespace sco;
 typedef BasicArray<Var> VarArray;
 
-struct CartPoseErrCalculator : public TrajOptVectorOfVector
+/**
+ * @brief The CartPosErrCalculator Used to calculate the error for CartPosTermInfo
```
Levi-Armstrong: I recommend removing `The CartPoseErrCalculator`

Levi-Armstrong: I recommend this for the others.

trajopt/include/trajopt/problem_description.hpp
```diff
@@ -166,112 +166,162 @@ struct TRAJOPT_API ProblemConstructionInfo
 };
 
 /** @brief This is used when the goal frame is not fixed in space */
-struct PoseCostInfo : public TermInfo, public MakesCost, public MakesConstraint
+struct CartPosTermInfo : public TermInfo, public MakesCost, public MakesConstraint
 {
+  /// Timestep at which to apply term
```
Levi-Armstrong:
```suggestion
  /** @brief Timestep at which to apply term */
```

Levi-Armstrong: Recommend the same for all others.

trajopt/include/trajopt/kinematic_terms.hpp
mpowelson: I agree. I had considered renaming it to DynamicCartPoseErrCalculator or something similar.  I'll remove the current CartPos and drop the Static on the other one.

mpowelson:
> I like the changes. I looked into it and the `MakeCost` and `MakeConstraint` are empty structures and I remember reading in one of the todo comments in the trajopt code a while back that the `TermInfo` was to replace these. I recommend as part of this effort to remove these classes.

I saw that those were empty, but I figured they were just there for future implementation. I can drop them.

> Also, I noticed that in class name containing `Pose` it was abbreviated to `Pos`. Is there a specific reason for the change?

The main reason was that we had StaticPoseCostInfo (cartesian pose) and JointPosCostInfo (presumably joint position). I figured if we abbreviate both to Pos, it will be more consistent since I would say a Cartesian Position and a Cartesian Pose are more or less interchangable in this case. I guess it was intentional ambiguity.

mpowelson:
@Levi-Armstrong If I remove MakesConstraint and MakesCost, what do you recommend I do with Trajopt/problem_description.cpp lines 150 and 174
`    if (!dynamic_cast<MakesConstraint*>(term.get()))
      PRINT_AND_THROW(boost::format("%s is only a cost, but you listed it as a constraint") % type);`

I don't believe we currently have any other way of checking if the term is capable of being a cost/constraint. As far as I can tell, that is the only place they are used.

Levi-Armstrong:
On the removing class cost/constraint lets save this for the next PR. I have an idea on how to address and I will test it once this is merged.

trajopt_examples/config/basic_cartesian_plan.json
```diff
@@ -11,7 +11,9 @@
       "type" : "joint_vel",
       "params":
       {
-        "coeffs" : [5]
+        "joint_name" : "all_joints",
+        "coeffs" : [5],
+        "penalty_type" : "SQUARED"
```
Levi-Armstrong: Just curious, but when would you use the other penalty types?

trajopt/src/problem_description.cpp
```diff
@@ -11,6 +11,7 @@
 #include <trajopt_utils/eigen_conversions.hpp>
 #include <trajopt_utils/eigen_slicing.hpp>
 #include <trajopt_utils/logging.hpp>
+#include <ros/ros.h>
```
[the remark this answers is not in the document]
mpowelson: I've been adding in ros console messages to inform the user when we are inferring information or something is wrong. e.g. if they give only one coefficient, I set all the joints to that value and warn.

trajopt/src/problem_description.cpp
```diff
@@ -517,69 +513,145 @@ void CartVelCntInfo::fromJson(ProblemConstructionInfo& pci, const Value& v)
...
+    for (int iStep = first_step; iStep < last_step; ++iStep)
+    {
+      prob.addConstraint(ConstraintPtr(new TrajOptConstraintFromErrFunc(
+          VectorOfVectorPtr(new CartVelErrCalculator(prob.GetKin(), prob.GetEnv(), link, max_displacement)),
+          MatrixOfVectorPtr(new CartVelJacCalculator(prob.GetKin(), prob.GetEnv(), link, max_displacement)),
+          concat(prob.GetVarRow(iStep), prob.GetVarRow(iStep + 1)),
+          VectorXd::Ones(0),
+          INEQ,
+          "CartVel")));
+    }
```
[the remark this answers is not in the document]
mpowelson: Just added

trajopt_examples/config/basic_cartesian_plan.json
mpowelson: This will be addressed in another PR where we integrate the tolerances into the joint costs

mpowelson:
@Levi-Armstrong Just resolved all of the changes. It builds on my machine. "Should" be good to go.

arocchi:
wanted to chime in regarding the renaming of `CartPose` to `CartPos`. Please notice they suggest two different things altogether. a `CartPos` would be a vector in `R³` as a `CartPose` would be in `SE(3)` (hence in particular an `Eigen::Isometry3d`). There might be cases where we want only to specify a `CartPos` constraint (e.g. free orientation), and times when we want to specify a `CartPose` constraint. I think keeping the naming clear will help users.

trajopt/include/trajopt/problem_description.hpp
```diff
@@ -166,112 +166,162 @@ struct TRAJOPT_API ProblemConstructionInfo
...
+  /** @brief For TT_COST: coefficient that scales cost. For TT_CNT: Velocity limit */
   DblVec coeffs;
+  /** @brief First time step to which the term is applied */
+  int first_step;
+  /** @brief Last time step to which the term is applied */
+  int last_step;
+  /** @brief Joint name to which the term is applied */
+  string joint_name;
+  /** @brief Penalty type applied */
+  PenaltyType penalty_type;
```
Levi-Armstrong: @mpowelson Based on our discussion was this supposed to removed?

Levi-Armstrong:
@mpowelson Some test are failing with the name changes. Can you please fix?

PR #38, opened 2018-11-05
Add exprMult(AffExpr, AffExpr)

mpowelson:
I won't pretend that I am an expert, but I think this is what we want. Based on the exprSquare. It does not consider any "shortcuts" if variables are repeated, but a) I am not sure how that would work and b) for our current usecase (multiplying by var=1/t) there would not be a repeated var anyway.

### DIFF
--- a/trajopt_sco/include/trajopt_sco/expr_ops.hpp
+++ b/trajopt_sco/include/trajopt_sco/expr_ops.hpp
@@ -151,6 +151,11 @@ inline QuadExpr exprSub(QuadExpr a, const QuadExpr& b)
 }
 
 //////////////////////
+/**
+ * @brief Multiplies two AffExpr. Does not consider any optimizations for shared variables
+ * @return The QuadExpr result of the multiplication
+ */
+QuadExpr exprMult(const AffExpr&, const AffExpr&);
 
 QuadExpr exprSquare(const Var&);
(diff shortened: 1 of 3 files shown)


trajopt_sco/include/trajopt_sco/expr_ops.hpp
```diff
@@ -151,6 +151,11 @@ inline QuadExpr exprSub(QuadExpr a, const QuadExpr& b)
...
 
 //////////////////////
+/**
+ * @brief Multiplies two AffExpr. Does not consider any optimizations for shared variables
+ * @return The QuadExpr result of the multiplication
+ */
+QuadExpr exprMult(const AffExpr&, const AffExpr&);
 
 QuadExpr exprSquare(const Var&);
 QuadExpr exprSquare(const AffExpr&);
```
Levi-Armstrong: To simplify things should we remove the exprSquare since it is a spacial case of exprMult?

mpowelson: We could, but I think the square probably gives you better performance if you know it is a square. It uses that knowledge to combine the like terms. E.g. (a+5)(a+5) = aa+10a+25 where multiply gives you (a+5)(a+5)=aa+5a+5a+25.

That being said, it may be simplified somewhere else in the optimizer and make no difference. But if so, I don't know where.

PR #40, opened 2018-11-09
Adds osqp and qpOASES solver interfaces

arocchi:
Adds osqp and qpOASES in the list of available solvers.
Automatically runs all tests that use a convex solver with all the list of solvers detected as available on the machine.
Disabled test calling solver with a non positive semi-definite matrix.
Please notice this needs a good refactor to simplify code in common between osqp and qpOASES solver, but at the moment there are issues with memory alignment that can cause weird errors whenever a serious refactor is attempted. All Eigen instances possibly causing alignment issues in the tesseract/trajopt stack should be reviewed.

Please review @Levi-Armstrong.


Levi-Armstrong:
@arocchi This is awesome and thank you for the contribution. I am on vacation next week and will try to get it reviewed this weekend.

Levi-Armstrong:
Also I submitted PR #42 that should fix the eigen aligned issues. For some reason I thought this was no longer need. It is possible that I did not catch all the places where it is need. I am unable to recreate the issue @mpowelson was having with the examples.

Levi-Armstrong:
I gave it a brief glance. I always hate asking because it seems nitpicky, but I plan to eventually migrate the style to using the ros cpp sytle guide so with new addition I would like to follow the ros style guide if that is not to painful. To help with formatting the repository has s a clang file. 

Also I notice the use of raw pointers a few places would it be good to replace them with shared pointers?

arocchi:
perfect, I'm on it!

arocchi:
refactored to take into account clang & roscpp guidelines. Doxygen is still lacking, but after we fix memory alignment issues and the solvers are refactored, cleanup + doxygen will follow

Levi-Armstrong:
@arocchi In search for the Eigen Issues I did some cleaning in trajopt packages and found one more structure without the Eigen Macro. I am unable to recreate the issue on my machine but let me know if you still have the issue if you test with PR #43.

arocchi:
Thanks @Levi-Armstrong - I will check  it later today and get back to you.

arocchi:
Hi @Levi-Armstrong, I tried #43 and the latest tesseract, still I get very weird results.

If you try this last commit, and run:
`trajopt_sco-test --gtest_filter=solver_utils.var_to_triplets`
you will see a really weird failure. Could you please take a look and confirm this happens also for you, and @mpowelson? The cpp code in solver_utils.cpp seems fine, still the behaviour of `std::vector` appears to be broken. I hope this is something I am doing wrong, but I suspect this is still related to the Eigen issue. Six eyes are better than two. If we are able to fix this, the solvers interface will be definitely cleaner and it will become much easier to add other solvers in the future.

@bhaskara if you could test this with latest tesseract (https://github.com/ros-industrial-consortium/tesseract/commit/f5eab0b87ba2e82c9f21b2c75cb89f79d24ccf7b) and confirm the test fails for you also it would be great. Thanks!

arocchi:
Tried code on new tesseract and trajopt, and find the issue (on my side that was causing the problem)
The new `solver_utils` work and I am in the process of refactoring the osqp and qpoases solver interfaces using them

arocchi:
@Levi-Armstrong @bhaskara please review. This refactors the new solvers using a shared set of `solver_utils` which is fairly clean, tested and allows to quickly add new solvers beyond `qpOASES` and `osqp`.

Levi-Armstrong:
@arocchi Do you mind re-basing on the latest and I will review it tomorrow?

arocchi:
will do!

arocchi:
@Levi-Armstrong I merged instead of rebasing, hope this is ok. Let me know otherwise! It is good to review. I added a new functionality to specify a solver in JSONs / `ProblemConstructionInfo`, I think this should need more testing. Please go ahead with the review, I will address comments and come up with more unit testing for this feature as I receive the review.

Levi-Armstrong:
There still appears to be conflicts.

arocchi:
As new branches are merged, conflicts appear. In particular, the last one was over new-lines. Fixed now.

trajopt_utils/src/logging.cpp
```diff
@@ -18,7 +18,11 @@ int LoggingInit()
     std::printf("You can set logging level with TRAJOPT_LOG_THRESH. Valid values: "
                 "%s. Defaulting to INFO\n",
                 VALID_THRESH_VALUES);
-    lvlstr = "INFO";
+  #ifdef NDEBUG
```
Levi-Armstrong: Why ERROR, DEBUG instead of DEBUG, INFO?

trajopt_sco/include/trajopt_sco/osqp_interface.hpp
```diff
@@ -0,0 +1,59 @@
...
+  OSQPSettings osqp_settings_;     // OSQP Settings
+  OSQPData osqp_data_;             // OSQPData
+  OSQPWorkspace* osqp_workspace_;  // OSQP Workspace
+
+  void updateObjective();
+  void updateConstraints();
+  void createOrUpdateSolver();
+
+public:
+  VarVector vars_;
```
Levi-Armstrong: Documentation for member variables would be helpful for other developers in the future.

trajopt_sco/include/trajopt_sco/osqp_interface.hpp
```diff
@@ -0,0 +1,59 @@
...
+  std::vector<c_int> A_column_pointers_;
+  DblVec A_csc_data_;
+  DblVec l_, u_;
+
+  QuadExpr objective_;
+
+  OSQPModel();
+  virtual ~OSQPModel();
+
+  Var addVar(const std::string& name);
```
Levi-Armstrong: override keyword is missing

trajopt_sco/include/trajopt_sco/qpoases_interface.hpp
```diff
@@ -0,0 +1,74 @@
...
+   */
+  bool updateSolver();
+
+  /**
+   * Instantiates a new qpOASES problem
+   */
+  void createSolver();
+
+public:
+  VarVector vars_;
```
Levi-Armstrong: Same as above about documentation for member variables.

trajopt_sco/include/trajopt_sco/qpoases_interface.hpp
```diff
@@ -0,0 +1,74 @@
...
+  IntVec A_column_pointers_;
+  DblVec A_csc_data_;
+  DblVec lbA_, ubA_;
+
+  QuadExpr objective_;
+
+  qpOASESModel();
+  virtual ~qpOASESModel();
+
+  Var addVar(const std::string& name);
```
Levi-Armstrong: same as above, override keyword missing for virtual methods.

Levi-Armstrong:
I am also looking into if we should use Factory style approach for the solvers.

Levi-Armstrong:
Overall looks great just a few minor things.

trajopt_sco/include/trajopt_sco/osqp_interface.hpp
arocchi: Actually in many cases these variables could safely be private. Should we document them also in those cases?

trajopt_sco/include/trajopt_sco/solver_interface.hpp
```diff
@@ -148,5 +149,53 @@ std::ostream& operator<<(std::ostream&, const Cnt&);
 std::ostream& operator<<(std::ostream&, const AffExpr&);
 std::ostream& operator<<(std::ostream&, const QuadExpr&);
 
-ModelPtr createModel();
+class ConvexSolver
+{
+public:
+  enum Value {
```
bhaskara: A name like `ModelType` seems to describe this better.

trajopt_sco/src/osqp_interface.cpp
```diff
@@ -0,0 +1,270 @@
...
+void OSQPModel::removeCnts(const CntVector& cnts)
+{
+  IntVec inds = cnts2inds(cnts);
+  for (unsigned i = 0; i < cnts.size(); ++i)
+    cnts[i].cnt_rep->removed = true;
+}
+
+void OSQPModel::updateObjective()
+{
+  int n = vars_.size();
```
bhaskara: Suggest making declarations `const` unless they actually change.

bhaskara: Also, this should presumably be `size_t`.

trajopt_sco/include/trajopt_sco/solver_interface.hpp
```diff
@@ -148,5 +149,53 @@ std::ostream& operator<<(std::ostream&, const Cnt&);
 std::ostream& operator<<(std::ostream&, const AffExpr&);
 std::ostream& operator<<(std::ostream&, const QuadExpr&);
 
-ModelPtr createModel();
+class ConvexSolver
```
bhaskara: Do we actually need a whole class for this given that it seems to just wrap an enum?

trajopt_sco/include/trajopt_sco/expr_ops.hpp
```diff
@@ -161,5 +160,4 @@ QuadExpr exprSquare(const Var&);
 QuadExpr exprSquare(const AffExpr&);
 
 AffExpr cleanupAff(const AffExpr&);
-QuadExpr cleanupQuad(const QuadExpr&);  // warning: might make it non-psd!
```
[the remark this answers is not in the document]
arocchi: this function was (luckily) not used and potentially very dangerous, as it changed the structure of the matrix with a very simplistic filtering. In general, we can "sparsify" a matrix by using a tolerance, but the tolerance used here was very high, much higher than numeric precision! In the future these things could be handled directly by the `expToEigen` methods - i.e., being able to select a threshold for sparsification

trajopt_sco/src/osqp_interface.cpp
```diff
@@ -0,0 +1,270 @@
...
+
+Cnt OSQPModel::addIneqCnt(const AffExpr& expr, const std::string& /*name*/)
+{
+  cnts_.push_back(new CntRep(cnts_.size(), this));
+  cnt_exprs_.push_back(expr);
+  cnt_types_.push_back(INEQ);
+  return cnts_.back();
+}
+
+Cnt OSQPModel::addIneqCnt(const QuadExpr&, const std::string& /*name*/)
```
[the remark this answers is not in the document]
arocchi: I substituted the assert and return with a throw, so that this happens clearly whenever used. At the moment quadratic inequality constraints are never used, and at the moment, among the supported solvers, only `Gurobi` supports them. If necessary, we might add support for QCQP  in the future. I can't come up at the moment with a more elegant solution at the moment.. ideas?

trajopt_sco/src/solver_interface.cpp
```diff
@@ -28,7 +36,7 @@ void simplify2(IntVec& inds, DblVec& vals)
   Int2Double ind2val;
   for (unsigned i = 0; i < inds.size(); ++i)
   {
-    if (vals[i] != 0)
+    if (vals[i] != 0.0)
```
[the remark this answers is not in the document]
arocchi: All these checks filter out only `0.0` for sparsification. I think any other "filtering" when doing sparsification is better avoided at this moment, and if needed in the future it should be done just in one place, possibly via a parameter, e.g. in `exprToEigen`

trajopt_utils/src/logging.cpp
arocchi: I basically tried to tune down the verbosity by default, from the very verbose "INFO" to "ERROR". If you prefer to keep it more verbose I can change it back!

trajopt_sco/include/trajopt_sco/solver_interface.hpp
arocchi: This was a solution I came up to have a place to easily manage conversions from string and to string, for logging and loading of the enum from json (strings are more readable than ints!).
I am open for suggestions in case we can make it more elegant!

trajopt_sco/include/trajopt_sco/solver_interface.hpp
arocchi: Thanks for the review! Yes, I am actually also a bit confused of the meaning of model vs solver that is used in the codebase. With the addition of solver_utils, I also have the feeling that all these different "Models" could be encapsulated in a single `Model`, which then make use of different Convex Solvers to solve them. For instance, the Model could provide Eigen types for the objective function and constraints, and the solver would use them as best as possible. Well, these suggestions of mine are probably not for this PR. I can see the idea beind the fact that `ModelType` would be more fitting here though. What do you think @Levi-Armstrong?

arocchi:
@Levi-Armstrong bump

trajopt_sco/include/trajopt_sco/osqp_interface.hpp
Levi-Armstrong: I think would be used for debugging or adding features later on for those not familiar.

trajopt_utils/src/logging.cpp
Levi-Armstrong: Sounds good. If not set it will default to error but if built in debug it will be set to debug.

Levi-Armstrong: Sound good.

trajopt_sco/include/trajopt_sco/solver_interface.hpp
Levi-Armstrong: If model is used throughout then probably best to keep consistent.

trajopt_sco/include/trajopt_sco/solver_interface.hpp
Levi-Armstrong: I plan to look at something similar to what is shown [here](http://www.dorodnic.com/blog/2014/03/29/abstract-factory/)  but this can be done in a separate PR.

Levi-Armstrong:
@arocchi Then only think left is the discussion around naming of enum to ModelType.

trajopt_sco/include/trajopt_sco/solver_interface.hpp
arocchi: on it

arocchi:
@Levi-Armstrong I renamed `ConvexSolver` into `ModelType`,  let me know your thoughts about it so that we can proceed with the PR

Levi-Armstrong:
@arocchi It is saying there are conflicts with the base branch. Do yo mind rebasing and then I will merge?

Levi-Armstrong:
@arocchi It will not let me merge it. It says `This branch cannot be rebased due to conflicts` but it does not give me any indication what the conflict is.

PR #47, opened 2018-11-29
Update JointPos terms 

mpowelson:
It looks like this currently breaks some tests, but I don't think I'll be able to get to look at it until next week, so I thought I would go ahead and push it. It could just be that the json needs to be updated.


mpowelson:
@Levi-Armstrong  This should be ready for review. I just had to add timesteps to the json files and change vals to targets.

trajopt/src/trajectory_costs.cpp
```diff
@@ -18,30 +18,227 @@ namespace trajopt
...
+  out << diff1, diff2;
+  return util::toDblVec(out);
+}
+
+sco::ConvexConstraintsPtr JointPosIneqConstraint::convex(const DblVec& /*x*/, sco::Model* model)
+{
+  sco::ConvexConstraintsPtr out(new sco::ConvexConstraints(model));
+  // Add hinge cost. Set the coefficient to 1 here since we include it in the AffExpr already
+  // This is necessary since we want a seperate coefficient per joint
+  for (sco::AffExpr expr : expr_vec_)
```
Levi-Armstrong: Same as above?
```suggestion
  for (sco::AffExpr& expr : expr_vec_)
```

trajopt/src/trajectory_costs.cpp
```diff
@@ -18,30 +18,227 @@ namespace trajopt
...
+  // Takes diff b/n the subsequent rows and subtract each row by the targets_ vector
+  Eigen::MatrixXd diff =
+      (traj.block(first_step_, 0, last_step_ - first_step_ + 1, traj.cols())).rowwise() - targets_.transpose();
+  // Squares it, multiplies it by a diagonal matrix of coefficients, and converts to vector
+  return util::toDblVec((diff.array().square()).matrix() * coeffs_.asDiagonal());
+}
+sco::ConvexConstraintsPtr JointPosEqConstraint::convex(const DblVec& /*x*/, sco::Model* model)
+{
+  sco::ConvexConstraintsPtr out(new sco::ConvexConstraints(model));
+  for (sco::AffExpr expr : expr_vec_)
```
Levi-Armstrong: Should we be making copies?
```suggestion
  for (sco::AffExpr& expr : expr_vec_)
```

trajopt/src/trajectory_costs.cpp
mpowelson: I also fixed in vel, acc, and jerk

mpowelson:
@Levi-Armstrong Made those changes

mpowelson:
For posterity: This is a breaking change. Most notably, val has been renamed to targets and timestep is now first_step and last_step. This allows the same position vector to be applied to multiple timesteps at once and mirrors the interface for vel, accel, and jerk.

PR #49, opened 2018-12-06
Remove the default target = 0 from joint cost/constraints and add default coeff = 1

mpowelson:
The idea is that we want at least one parameter that must be set for each cost/cnt. It didn't seem to make much sense that coeff was that value. Explicitly setting the target and letting coeff default to 1 seemed like a better idea.


Levi-Armstrong:
This can be merged once PR #47  is merged.

Levi-Armstrong:
@mpowelson Can you rebase?

mpowelson:
@Levi-Armstrong Done. I also fixed the json in the two examples. I didn't realize before that I broke them. This is a breaking change if your terms didn't have a target define before and were relying on the default.

PR #50, opened 2018-12-07
Integrate time param - term type switch

mpowelson:
Time param sections are still skeletons. But it doesn't appear to have broken anything without time.

PR #51, opened 2018-12-10
Integrate time param

mpowelson:
Only joint velocity and total time have been integrated. Some notes

1) Time param costs are written to be applied to a single joint. In order to keep the interface consistent, I am currently just looping over them (This will be visible to the user). These should probably be revisited at some point.


mpowelson:
@Levi-Armstrong This is ready to review if want. I'm going to try to get the cartesian costs integrated, but this code should be complete as is. I think I'll add those in a separate PR to keep things somewhat manageable.

Levi-Armstrong:
@mpowelson PR #50 has been merged, please rebase

mpowelson:
Just rebased it. On my machine all of the tests pass except the octomap which fails because of the Eigen issue. This shouldn't break anything in the non time usage, but there is still the question of whether or not the behavior we were getting with the hinge cost/inequality constraint was a bug or just the nature of the optimization.

trajopt/include/trajopt/problem_description.hpp
```diff
@@ -358,7 +360,8 @@ struct JointVelTermInfo : public TermInfo
   void hatch(TrajOptProb& prob) override;
   DEFINE_CREATE(JointVelTermInfo)
 
-  JointVelTermInfo() : TermInfo(TT_COST | TT_CNT) {}
+  /** @brief Initialize term with it's supported types */
+  JointVelTermInfo() : TermInfo(TT_COST | TT_CNT | TT_USE_TIME) {}
```
Levi-Armstrong: should it be this way? Issue it is not possible to just have the term type as TT_USE_TIME.
```suggestion
   JointVelTermInfo() : TermInfo(TT_COST | TT_CNT | (TT_COST + TT_USE_TIME)| ( TT_CNT + TT_USE_TIME)) {}
```

trajopt/include/trajopt/problem_description.hpp
```diff
@@ -486,4 +491,21 @@ struct CollisionTermInfo : public TermInfo
...
+  /** @brief Scales this term. */
+  double coeff = 1.0;
+  /** @brief If non zero, penalty type will be a hinge on values greater than this limit relative to the target. */
+  double limit = 0.0;
+
+  void hatch(TrajOptProb& prob);
+  void fromJson(ProblemConstructionInfo &pci, const Json::Value &v);
+  DEFINE_CREATE(TotalTimeTermInfo)
+
+  TotalTimeTermInfo(): TermInfo(TT_COST | TT_CNT | TT_USE_TIME) {}
```
Levi-Armstrong: same as above

trajopt/src/problem_description.cpp
```diff
@@ -683,7 +684,7 @@ void CartVelTermInfo::fromJson(ProblemConstructionInfo& pci, const Json::Value&
 
 void CartVelTermInfo::hatch(TrajOptProb& prob)
 {
-    unsigned int n_dof = prob.GetKin()->numJoints();
+  unsigned int n_dof = prob.GetKin()->numJoints();
 
   if (term_type == (TT_COST | TT_USE_TIME))
```
Levi-Armstrong:
```suggestion
   if (term_type == (TT_COST  +  TT_USE_TIME))
```

trajopt/test/costs_unit.cpp
```diff
@@ -463,12 +467,227 @@ TEST_F(CostsTest, inequality_jointVel)
...
+  jv->name = "joint_vel_limits";
+  jv->term_type = TT_CNT | TT_USE_TIME;
+  pci.cnt_infos.push_back(jv);
+
+  // Joint Velocities also have a cost to some non zero value
+  std::shared_ptr<JointVelTermInfo> jv2 = std::shared_ptr<JointVelTermInfo>(new JointVelTermInfo);
+  jv2->coeffs = std::vector<double>(7, 1.0);
+  jv2->targets = std::vector<double>(7.0, cost_targ1);
+  //  jv2->lower_tols = std::vector<double>(7.0, -0.01);
+  //  jv2->upper_tols = std::vector<double>(7.0, 0.0);
```
Levi-Armstrong: Should this be commented out?

trajopt/test/costs_unit.cpp
```diff
@@ -463,12 +467,227 @@ TEST_F(CostsTest, inequality_jointVel)
...
+  jv2->last_step = (pci.basic_info.n_steps - 1) / 2;
+  jv2->name = "joint_vel_targ_1";
+  jv2->term_type = TT_COST | TT_USE_TIME;
+  pci.cost_infos.push_back(jv2);
+
+  std::shared_ptr<JointVelTermInfo> jv3 = std::shared_ptr<JointVelTermInfo>(new JointVelTermInfo);
+  jv3->coeffs = std::vector<double>(7, 1.0);
+  jv3->targets = std::vector<double>(7.0, cost_targ2);
+  //  jv3->lower_tols = std::vector<double>(7.0, -0.01);
+  //  jv3->upper_tols = std::vector<double>(7.0, 0.01);
```
Levi-Armstrong: should this be commented out.

trajopt/include/trajopt/problem_description.hpp
mpowelson: These should be equivalent. 001 | 010 | 100 = 111 and 001 | 010 | 101 | 110 = 111. 

Now another issue that does occur to me is that it is currently a bit ambiguous whether a cost that does not have time is set that way or just hasn't been set yet.

trajopt/src/problem_description.cpp
mpowelson: These should be equivalent as well. 
https://stackoverflow.com/questions/7334832/are-addition-and-bitwise-or-the-same-in-this-case

trajopt/test/costs_unit.cpp
mpowelson: No. I'll fix that. It should behave about the same regardless.

trajopt/src/problem_description.cpp
mpowelson: That being said, if you'd rather me go with + instead of |, I can change them all to be that way. I've just always seen people use masks with and/or

mpowelson:
@Levi-Armstrong Just pushed fixed jacobian

mpowelson:
@Levi-Armstrong As we were discussing, I actually do have the Acceleration and Jerk code in there for dt. However, JointAccel and JointJerk don't have TT_USE_TIME set, so if someone tries to use them it should error.

Still TODO is to convert accel and jerk to use 1/dt.

mpowelson:
@Levi-Armstrong What is wrong with this melodic build? It looks like it timed out?

mpowelson:
@Levi-Armstrong This is ready for another look if you get some time.

PR #52, opened 2018-12-10
Add additional compiler warning options

Levi-Armstrong:


trajopt_utils/src/clock.cpp
```diff
@@ -29,8 +33,8 @@ double GetClock()
 {
   struct timeval startTimeStruct;
   unsigned long int curTime;
-  gettimeofday(&startTimeStruct, NULL);
-  curTime = startTimeStruct.tv_sec * (long unsigned int)(1e6) + startTimeStruct.tv_usec;
-  return (1e-6) * (curTime - startTime);
+  gettimeofday(&startTimeStruct, nullptr);
+  curTime = static_cast<unsigned long>(startTimeStruct.tv_sec * 1e6l + startTimeStruct.tv_usec);
```
mpowelson: Is this supposed to be 1e6**1**?

Levi-Armstrong: That is interesting GitHub make it look like the number one but it is actually the letter l (Long).

mpowelson: Ah. That makes sense. Then I'd say it's good to merge.

PR #59, opened 2019-01-03
Add pick_and_place_plan example

mpowelson:
Adds an example that uses the Tesseract Planner interface to perform a pick and place task with the KUKA iiwa 7kg. 

Notes:
1. This includes the URDF from the [IFL-CAMP iiwa driver repo](https://github.com/IFL-CAMP/iiwa_stack/tree/master/iiwa_description). I added its license (BSD) to the package.xml and made a note in the readme, but I know this is still something we might not want to do.
2) This obviously depends on Tesseract. It could perhaps be a Tesseract example instead. This is currently more convenient as there is no tesseract_examples package.
3) I added an rviz config for this example. I thought it would be nice to have an example that was literally just calling one launch file.


mpowelson:
I rebased this to support #51

mpowelson:
@Levi-Armstrong I rebased again to fix the conflicts. I could be pursuaded to remove the last commit if you'd rather keep in non time param.

mpowelson:
So something was seriously broken in the commit history. I guess I was pretty liberal in my rebasing of the time param branch. Anyway, I blew away the history, cherry-picked the commits, and here we are.

trajopt_examples/urdf/iiwa7/iiwa7.xacro
```diff
@@ -0,0 +1,333 @@
+<?xml version="1.0"?>
```
Levi-Armstrong: I would add the BSD license at the top of the file.

Levi-Armstrong:
looks like it clang failed

PR #73, opened 2019-02-07
Bug fixes for examples

mpowelson:
Fixes #55 and a few other things that I have found along the way.

PR #74, opened 2019-02-12
Example unit test

mpowelson:
Adds unit tests to run all of the examples. Depends on #73 . 

Still WIP because car_seat_demo is broken (see [tesseract issue 71](https://github.com/ros-industrial-consortium/tesseract/issues/71#)).

### DIFF
--- a/dependencies.rosinstall
+++ b/dependencies.rosinstall
@@ -2,3 +2,8 @@
     local-name: tesseract
     uri: https://github.com/ros-industrial-consortium/tesseract.git
     version: kinetic-devel
+
+- git:
+    local-name: motoman
+    uri: https://github.com/ros-industrial/motoman.git
+    version: kinetic-devel
--- a/trajopt_examples/CMakeLists.txt
+++ b/trajopt_examples/CMakeLists.txt
@@ -3,6 +3,8 @@ project(trajopt_examples)
 
 add_compile_options(-std=c++11 -Wall -Wextra)
 
+find_package(rostest REQUIRED)
+
 find_package(catkin REQUIRED COMPONENTS
   roscpp
   roslib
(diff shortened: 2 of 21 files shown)


mpowelson:
@Levi-Armstrong Also note that glass upright and basic cartesian are are hardcoded to use the json method. How should I address that?

Levi-Armstrong:
What is the current issue with just using json method?

mpowelson:
If we break something in the C++ interface at runtime this won't catch it. I could add it as a ROS param, but I was not sure how to best include that in the test. I suppose I could just launch the node twice, one with each param value.

Levi-Armstrong:
You should be able to pass arguments to the launch file as shown below. You could add two test in the CMakeLists.txt passing an argument telling it to use json file or c++.
```
add_rostest(test/testlaunchfile.test ARGS argument1:=vaule1 argument2:=value2)
```

mpowelson:
I didn't do that because I am using add_rostest_gtest which seems to make an executable? I didn't want to have 2 executables for each. Maybe I have this structured inefficiently. I just launched 2 different test nodes in the launch file.

mpowelson:
@Levi-Armstrong I forgot that we need to add the motoman driver to CI or else the car seat demo is going to fail. How should I go about that?

mpowelson:
@Levi-Armstrong Do you want to merge this into kinetic-devel or should I close this and open one against Tesseract? Or is this already done in Tesseract?

Regardless, I think the cmake method of running these might be cleaner anyway.

Levi-Armstrong:
Open against tesseract?

mpowelson:
Didn't all of the examples get moved to tesseract?

PR #78, opened 2019-02-14
Updates to account for changes in tesseract_collision

Levi-Armstrong:


Levi-Armstrong:
https://github.com/ros-industrial-consortium/tesseract/pull/72

PR #79, opened 2019-02-14
removed dependency on pcl_conversions so that pcl_1.8 installs compile

drchrislewis:
ran a clean and then a build, all went well. Not sure what would happen if only pcl1.7 is installed

### DIFF
--- a/trajopt/CMakeLists.txt
+++ b/trajopt/CMakeLists.txt
@@ -78,7 +78,7 @@ install(DIRECTORY include/${PROJECT_NAME}/
 if (CATKIN_ENABLE_TESTING)
 
   find_package(rostest REQUIRED)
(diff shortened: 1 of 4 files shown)


Levi-Armstrong:
@drchrislewis Travis fails. It looks like a header should be remove shown below.

```
/root/catkin_ws/src/trajopt_ros/trajopt_examples/src/basic_cartesian_plan.cpp:35:45: fatal error: pcl_conversions/pcl_conversions.h: No such file or directory
```

drchrislewis:
@Levi-Armstrong perhaps you need to do a sudo apt-get install ros-kinetic-pcl-conversions on the docker so that /opt/ros/kinetic/include/pcl_conversions/pcl_conversions.h exists

Levi-Armstrong:
This was addressed when the package was converted to a pure cmake.

PR #89, opened 2019-02-27
pick_and_place example Improvements

mpowelson:
Fixes the indexing on the place move. Also allows steps_per_phase to be set in the launch file and adds 4 more place locations.


mpowelson:
@Levi-Armstrong I didn't realize this had failed clang too. Fixed manually again.

PR #100, opened 2019-03-29
Add raster_path_plan Example

mpowelson:
Adds example showing Tesseract Freespace and Tesseract Array planners. These are wrappers for TrajOpt to make it easier to use for some common use cases.


mpowelson:
@Levi-Armstrong Do you still want to merge this into kinetic-devel or should I just close this and open one against Tesseract?

PR #114, opened 2019-07-18
Fix calcCollision to filter link specific distance threshold

Levi-Armstrong:
This was an oversight on my part when I added the ability to add link specific distance threshold. The filtering was occurring in convex and value function but not in the calcCollision so cost function were being added for pairs that were outside the threshold. This was causing issues because cost returned from the value function and the sum of the constants from the cost function were not even close as a result. This was causing issues with the optimizer as you would expect.

Also this fixes an issues I ran into with a return type.

### DIFF
--- a/trajopt/include/trajopt/utils.hpp
+++ b/trajopt/include/trajopt/utils.hpp
@@ -73,7 +73,7 @@ struct SafetyMarginData
    * @param safety_margin_coeffs A safety margin coefficient vector where each
    * element corresponds to a given timestep.
    */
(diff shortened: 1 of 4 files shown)


Levi-Armstrong:
@mpowelson In the process of fixing the pick and place I found this issue. Do you mind reviewing the PR?

Levi-Armstrong:
@mpowelson I added documentation to the SafetyMarginData class and fixed a style guide issues with the naming of SetSafetyMarginData. This breaks one of the examples in tesseact, which I am adding to  my current tesseract PR.

PR #117, opened 2019-07-27
Add assert in kinematic_terms for null kinematic link

Levi-Armstrong:


trajopt/include/trajopt/kinematic_terms.hpp
```diff
@@ -79,6 +79,7 @@ struct CartPoseErrCalculator : public TrajOptVectorOfVector
     , tcp_(tcp)
   {
     kin_link_ = adjacency_map_->getLinkMapping(link_);
+    assert(kin_link_ != nullptr);
```
mpowelson: So this happens when the link is not in the kinematic chain?

That seems like a common enough mistake you might consider using an if and providing an error message. It should only effect ConstructProblem which I would think most people won't care about performance there.

Levi-Armstrong: I added an error message prior to the assert.

PR #125, opened 2019-08-27
Update DynamicCartPose to allow target link tcp

Levi-Armstrong:


Levi-Armstrong:
>  However won't you need to update the problem description for this to be useful to people? fromJson and hatch will both need to be changed.

Good catch. I updated the fromJson and hatch.

PR #128, opened 2019-09-09
Feature/add log files

Levi-Armstrong:
This add ability to log iteration results for plotting convergence of the solver, costs and constraints.

### DIFF
--- a/trajopt_sco/src/optimizers.cpp
+++ b/trajopt_sco/src/optimizers.cpp
@@ -3,6 +3,7 @@ TRAJOPT_IGNORE_WARNINGS_PUSH
 #include <boost/format.hpp>
 #include <cmath>
 #include <cstdio>
+#include <stdio.h>
 TRAJOPT_IGNORE_WARNINGS_POP
 
 #include <trajopt_sco/expr_ops.hpp>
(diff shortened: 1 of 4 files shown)


Levi-Armstrong:
@jnicho Could you review?

trajopt_sco/src/optimizers.cpp
```diff
@@ -271,11 +222,237 @@ struct MultiCritFilter {
...
+    std::fprintf(stream, ",%e", x);
+
+  std::fprintf(stream, "\n");
+}
+
+void BasicTrustRegionSQPResults::writeCosts(std::FILE* stream, bool header) const
+{
+  if (header)
+  {
+    std::fprintf(stream, "%s", "COST NAMES");
```
mpowelson: what would you think about writing the cost names 4 times each? The reason I say that is that when I pull the resulting log into excel, the costs are above some columns. But right now those aren't the columns they are associated with. If we printed them 4 times each then they would be above the oldexact, dappprox, etc columns that describe that cost.

Same for constraints obviously. I attached some logs for convenience
[trajopt_constraints.log](https://github.com/ros-industrial-consortium/trajopt_ros/files/3611350/trajopt_constraints.log)
[trajopt_costs.log](https://github.com/ros-industrial-consortium/trajopt_ros/files/3611351/trajopt_costs.log)
[trajopt_solver.log](https://github.com/ros-industrial-consortium/trajopt_ros/files/3611352/trajopt_solver.log)
[trajopt_vars.log](https://github.com/ros-industrial-consortium/trajopt_ros/files/3611353/trajopt_vars.log)

trajopt_sco/src/optimizers.cpp
```diff
@@ -271,11 +222,237 @@ struct MultiCritFilter {
...
+
+  std::fprintf(stream, "%s", "COSTS");
+  for (size_t i = 0; i < old_cost_vals.size(); ++i)
+  {
+    double approx_improve = old_cost_vals[i] - model_cost_vals[i];
+    double exact_improve = old_cost_vals[i] - new_cost_vals[i];
+    std::fprintf(
+        stream, ",%e,%e,%e,%e", old_cost_vals[i], approx_improve, exact_improve, exact_improve / approx_improve);
+  }
+  std::fprintf(stream, "\n");
```
mpowelson: It looks like we might want an fflush here. On my pc the data was not being written until the file was closed without it. That would mess up the live plotting code we had talked about.

Levi-Armstrong: Done.

trajopt_sco/src/optimizers.cpp
Levi-Armstrong: Good idea. Done.

Levi-Armstrong:
@mpowelson I made the requested changes. Are the changes to how the log file is written correct?

trajopt_sco/src/optimizers.cpp
```diff
@@ -271,11 +222,272 @@ struct MultiCritFilter {
...
+  {
+    std::fprintf(stream, "%s", "COST NAMES");
+    for (const auto& name : cost_names)
+      std::fprintf(stream, ",%s,%s,%s,%s", name.c_str(), name.c_str(), name.c_str(), name.c_str());
+
+    std::fprintf(stream, "\n");
+
+    std::fprintf(stream, "%s", "DESCRIPTION");
+    for (size_t i = 0; i < cost_names.size(); ++i)
+      std::fprintf(stream, ",%s,%s,%s,%s\n", "oldexact", "dapprox", "dexact", "ratio");
```
mpowelson: drop the \n here. Same in constraints

```
std::fprintf(stream, ",%s,%s,%s,%s", "oldexact", "dapprox", "dexact", "ratio");
```

mpowelson: > @mpowelson I made the requested changes. Are the changes to how the log file is written correct?

See the one comment. When I made that change it looked fine.

Levi-Armstrong: Should be fixed now.

PR #130, opened 2019-09-16
Cartesian Pose Term Info from frame relative transform

marip8:
This PR updates the Cartesian pose term info to be constructed from a transformation that is relative to an arbitrary frame that exists in the environment (rather than a transformation relative to the base frame of the kinematic chain)

### DIFF
--- a/trajopt/include/trajopt/problem_description.hpp
+++ b/trajopt/include/trajopt/problem_description.hpp
@@ -260,6 +260,8 @@ struct CartPoseTermInfo : public TermInfo
   std::string link;
   /** @brief Static transform applied to the link */
   Eigen::Isometry3d tcp;
+  /** @brief The frame relative to which the target position is defined */
+  std::string target;
 
   CartPoseTermInfo();
 
--- a/trajopt/include/trajopt/problem_description.hpp
+++ b/trajopt/include/trajopt/problem_description.hpp
@@ -221,6 +221,7 @@ struct DynamicCartPoseTermInfo : public TermInfo
 
   /** @brief Timestep at which to apply term */
   int timestep;
+  /** @brief The link relative to which the target_tcp is defined */
   std::string target;
   /** @brief Coefficients for position and rotation */
   Eigen::Vector3d pos_coeffs, rot_coeffs;
@@ -260,7 +261,8 @@ struct CartPoseTermInfo : public TermInfo
   std::string link;
   /** @brief Static transform applied to the link */
   Eigen::Isometry3d tcp;
-  /** @brief The frame relative to which the target position is defined */
+  /** @brief The frame relative to which the target position is defined. If empty, frame is assumed to the root,
+   * "world", frame */
   std::string target;
 
   CartPoseTermInfo();
--- a/trajopt/src/problem_description.cpp
+++ b/trajopt/src/problem_description.cpp
@@ -637,7 +637,8 @@ void DynamicCartPoseTermInfo::hatch(TrajOptProb& prob)
     catch (const std::exception&)
     {
       std::stringstream ss;
-      ss << "Failed to find transform for link '" + prob.GetKin()->getBaseLinkName() + "'; using identity transform instead";
+      ss << "Failed to find transform for link '" + prob.GetKin()->getBaseLinkName() +
+                "'; using identity transform instead";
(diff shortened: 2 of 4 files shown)


trajopt/include/trajopt/problem_description.hpp
```diff
@@ -259,6 +259,8 @@ struct CartPoseTermInfo : public TermInfo
   std::string link;
   /** @brief Static transform applied to the link */
   Eigen::Isometry3d tcp;
+  /** @brief The frame relative to which the target position is defined */
+  std::string target;
```
Levi-Armstrong: Might add a comment that if empty string the frame is assumed to be in world coordinate system.

trajopt/src/problem_description.cpp
```diff
@@ -601,7 +601,17 @@ void DynamicCartPoseTermInfo::hatch(TrajOptProb& prob)
...
+    try
+    {
+      world_to_base = state->transforms.at(prob.GetKin()->getBaseLinkName());
+    }
+    catch (const std::exception&)
+    {
+      std::stringstream ss;
+      ss << "Failed to find transform for link '" + prob.GetKin()->getBaseLinkName() + "'; using identity transform instead";
+      CONSOLE_BRIDGE_logError(ss.str().c_str());
+    }
```
Levi-Armstrong: What case did you run into that this was required? The kinematics object is created from the environment, so the base link of the kinematics should be in the state transforms.

trajopt/src/problem_description.cpp
```diff
@@ -673,7 +690,30 @@ void CartPoseTermInfo::hatch(TrajOptProb& prob)
...
+
+  Eigen::Isometry3d base_to_target = Eigen::Isometry3d::Identity();
+  try
+  {
+    base_to_target = state->transforms.at(target);
+  }
+  catch (const std::exception& ex)
+  {
+    std::stringstream ss;
+    ss << "Failed to find transform for link '" + prob.GetKin()->getBaseLinkName() + "'; using identity transform instead";
```
Levi-Armstrong: Should be `target` instead of `prob.GetKin()->getBaseLinkName()`

trajopt/src/problem_description.cpp
```diff
@@ -673,7 +690,30 @@ void CartPoseTermInfo::hatch(TrajOptProb& prob)
...
+    world_to_base = state->transforms.at(prob.GetKin()->getBaseLinkName());
+  }
+  catch(const std::exception&)
+  {
+    std::stringstream ss;
+    ss << "Failed to find transform for link '" + prob.GetKin()->getBaseLinkName() + "'; using identity transform instead";
+    CONSOLE_BRIDGE_logError(ss.str().c_str());
+  }
+
+  Eigen::Isometry3d base_to_target = Eigen::Isometry3d::Identity();
```
Levi-Armstrong: recommend renaming to `world_to_target`

trajopt/src/problem_description.cpp
marip8: In the case where someone accidentally provides the name of a link that doesn't exist (which I've done a few times on accident), the program should handle the exception thrown by `.at()` gracefully rather than dying

Levi-Armstrong: Since this is not recoverable at this state in the code, I think it would be best to propagate the exception. You could also check if the link exist first and if not use the trajopt PRINT_AND_THROW macro to throw an exception. This way if we ever added a gui on top it could propagate all the way up to the GUI where it is recoverable.

marip8: I changed this to the `PRINT_AND_THROW` as requested, but that has created a circular dependency with this related `tesseract` [PR](https://github.com/ros-industrial-consortium/tesseract/pull/125). If I make this change here, the `tesseract_motion_planners` unit tests provide an empty string to the `target` variable, resulting in an exception being thrown due to inability to find a link with no name.

Should I drop the commit where I implemented this, and open a new PR after it and [125](https://github.com/ros-industrial-consortium/tesseract/pull/125) get merged?

marip8: It looks like updating the `.travis.yml` to whitelist only `trajopt` packages solves this issue

PR #131, opened 2019-09-17
Add jacobian calculation to the cartesian position costs

mpowelson:
Previously, the cartesian costs were being numerically differentiated inside CostFromErrFunc. Now, the jacobian is being passed in as well.

See the comments in the code as well as https://github.com/mpowelson/trajopt_ros/pull/1 for more details, but it works as follows. 
1) The geometric jacobian is retrieved from the kinematics object (kdl currently).
2) The timestep is perturbed by a small amount and geometric jacobian is used to calculate the analytic jacobian

Additionally, unit tests where added that compare the old jacobian results (numerically calculated) to the new results.


mpowelson:
@Levi-Armstrong I removed the merge commit and rebased. I didn't see any drastic conflicts though. I only saw one spot where it didn't like the indices that was added.

PR #132, opened 2019-09-18
Add OptStatus to TrajOptResult

haudren:
I would like to propose the following change: add the `OptStatus` field to `TrajOptResult`. Right now, using `OptimizeProblem` is a useful convenience function that sets quite a few defaults automatically, but the result struct it returns is incomplete. Indeed, it's pretty much impossible to understand if the optimization succeeded or not based on only the trajectory, costs and constraints values.

Let me know what you think!

### DIFF
--- a/trajopt/include/trajopt/problem_description.hpp
+++ b/trajopt/include/trajopt/problem_description.hpp
@@ -85,6 +85,7 @@ struct TRAJOPT_API TrajOptResult
   std::vector<std::string> cost_names, cnt_names;
   DblVec cost_vals, cnt_viols;
   TrajArray traj;
+  sco::OptStatus status;
   TrajOptResult(sco::OptResults& opt, TrajOptProb& prob);
 };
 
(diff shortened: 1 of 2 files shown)


mpowelson:
@Levi-Armstrong Seems reasonable to me.

trajopt/include/trajopt/problem_description.hpp
```diff
@@ -85,6 +85,7 @@ struct TRAJOPT_API TrajOptResult
   std::vector<std::string> cost_names, cnt_names;
   DblVec cost_vals, cnt_viols;
   TrajArray traj;
+  sco::OptStatus status;
```
Levi-Armstrong: Since these are populated from the constructor should we make everything const in this struct?

haudren: I think it could, but I believe it's slightly outside the scope of that PR. I gave maintainer access anyways, feel free to add what you deem necessary, but I aimed at keeping it simple.

haudren:
I believe this is mergeable as-is, but let me know @Levi-Armstrong @mpowelson if you would like more improvements.

Issue #27, opened 2018-10-29
Basic cartesian example causes Eigen unaligned array assertion failure


mpowelson:
Fresh copy of the current kinetic-devel tesseract/trajopt gives me this error when I run the basic cartesian example. I've seen it for a while, but I thought it was only in the time param branch.

```
trajopt_examples_basic_cartesian_plan: /usr/include/eigen3/Eigen/src/Core/DenseStorage.h:128: Eigen::internal::plain_array<T, Size, MatrixOrArrayOptions, 32>::plain_array() [with T = double; int Size = 4; int MatrixOrArrayOptions = 0]: Assertion `(reinterpret_cast<size_t>(eigen_unaligned_array_assert_workaround_gcc47(array)) & (31)) == 0 && "this assertion is explained here: " "http://eigen.tuxfamily.org/dox-devel/group__TopicUnalignedArrayAssert.html" " **** READ THIS WEB PAGE !!! ****"' failed.
[trajopt_examples_basic_cartesian_plan-3] process has died [pid 860, exit code -6, cmd /home/mpowelson/workspaces/temp/devel/lib/trajopt_examples/trajopt_examples_basic_cartesian_plan __name:=trajopt_examples_basic_cartesian_plan __log:=/home/mpowelson/.ros/log/44dd890a-dbc5-11e8-aef7-8c1645997b24/trajopt_examples_basic_cartesian_plan-3.log].
```

gdb backtrace
```
#0  0x00007ffff3a66428 in __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:54
#1  0x00007ffff3a6802a in __GI_abort () at abort.c:89
#2  0x00007ffff3a5ebd7 in __assert_fail_base (fmt=<optimized out>, 
    assertion=assertion@entry=0x5cb7c8 "(reinterpret_cast<size_t>(eigen_unaligned_array_assert_workaround_gcc47(array)) & (31)) == 0 && \"this assertion is explained here: \" \"http://eigen.tuxfamily.org/dox-devel/group__TopicUnalignedArrayAss"..., 
    file=file@entry=0x5cb790 "/usr/include/eigen3/Eigen/src/Core/DenseStorage.h", line=line@entry=128, 
[...]

PR #133, opened 2019-09-20
Disable AVX Instructions to Fix Eigen Alignment Issues

mpowelson:
Fixes issue #27 

See https://github.com/ros-industrial-consortium/tesseract/issues/127 for more information


trajopt/CMakeLists.txt
```diff
@@ -45,6 +45,7 @@ if(CXX_FEATURE_FOUND EQUAL "-1")
 else()
     target_compile_features(${PROJECT_NAME} PUBLIC cxx_std_11)
 endif()
+target_compile_options(${PROJECT_NAME} PUBLIC -mno-avx)
```
gavanderhoorn: Does this work with compilers other than GCC?

Levi-Armstrong: It looks like this may work with Clang and AppleClang per [VTK](https://gitlab.kitware.com/vtk/vtk-m/commit/514ea09afcf6be56f887c96858d55a9a66fb37cb) but not all compilers. I did look and does no appear to be a feature so probably need to do something similar to what VTK did in the like. @gavanderhoorn do you know of any other solutions?

gavanderhoorn: No, not really.

I'm not sure we'd need something as elaborate as the VTK implementation, but if you need to set these compiler options across many different packages something which avoids duplication would be good.

We could compromise and only 'support' GCC for now by wrapping these `target_compile_options(..)` in a:

```cmake
if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
endif()
```

Levi-Armstrong: @gavanderhoorn We updated it with your suggestion and added a warning if not using GNU.

Issue #139, opened 2019-10-17
Using TCP causes issues in cartesian cost


mpowelson:
When I set the tip of the kinematic chain to be the frame that I want, it seems to solve fine. However when I set the tip of the chain to tool0 and then add a TCP that transforms it to the same position, it does not solve.

@Levi-Armstrong mentioned that this appeared after we updated the jacobian for those costs, so that might not be handled correctly.

PR #141, opened 2019-10-21
Disable jacobian calculator for cart pose and dynamic cart pose

Levi-Armstrong:
@mpowelson I am disabling the jacobian calculator for cart pose and dynamic cart pose. It appears to be intermittently providing incorrect results.


mpowelson:
This essentially reverts #131 due to a bug documented in #139. In short, the jacobian calculator does not appear to correctly handle TCPs

PR #146, opened 2019-12-02
Update interface to OSQP 0.6.0

haudren:
Since [version 0.6.0](https://github.com/oxfordcontrol/osqp/releases/tag/v0.6.0) OSQP's interface has changed. `osqp_setup` now returns an error code and fills in the workspace (which is now a pointer to pointer). This is the minimum change required to make this work, but I guess this PR is more a discussion:
- How should the deprecation / version change be handled?
- Properly taking into account the return code requires a number of tweaks to the interface, I'm not sure I understand the ramifications so I'd like some guidance.

### DIFF
--- a/trajopt_sco/src/osqp_interface.cpp
+++ b/trajopt_sco/src/osqp_interface.cpp
@@ -172,7 +172,7 @@ void OSQPModel::createOrUpdateSolver()
   if (osqp_workspace_ != nullptr)
     osqp_cleanup(osqp_workspace_);
   // Setup workspace - this should be called only once
-  osqp_workspace_ = osqp_setup(&osqp_data_, &osqp_settings_);
+  osqp_setup(&osqp_workspace_, &osqp_data_, &osqp_settings_);
 }
 
 void OSQPModel::update()
--- a/trajopt_sco/src/osqp_interface.cpp
+++ b/trajopt_sco/src/osqp_interface.cpp
@@ -177,7 +177,7 @@ void OSQPModel::createOrUpdateSolver()
     osqp_cleanup(osqp_workspace_);
   // Setup workspace - this should be called only once
   auto ret = osqp_setup(&osqp_workspace_, &osqp_data_, &osqp_settings_);
-  if(ret)
+  if (ret)
   {
     throw std::runtime_error("Could not initialize OSQP: error " + std::to_string(ret));
   }
--- a/trajopt_sco/src/osqp_interface.cpp
+++ b/trajopt_sco/src/osqp_interface.cpp
@@ -102,7 +102,11 @@ void OSQPModel::updateObjective()
 
   Eigen::SparseMatrix<double> sm;
   exprToEigen(objective_, sm, q_, n, true);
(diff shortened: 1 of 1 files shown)


Levi-Armstrong:
@haudren Thank you for the contribution. Is version 0.6.0 available on both kinetic and melodic? If so we can check the version in cmake and if less than 0.6.0 throw a cmake error. 

What changes to the interface would be required?

trajopt_sco/src/osqp_interface.cpp
```diff
@@ -102,7 +102,11 @@ void OSQPModel::updateObjective()
 
   Eigen::SparseMatrix<double> sm;
   exprToEigen(objective_, sm, q_, n, true);
-  eigenToCSC(sm, P_row_indices_, P_column_pointers_, P_csc_data_);
+
+  // Copy triangular upper into empty matrix
+  Eigen::SparseMatrix<double> triangular_sm;
+  triangular_sm = sm.triangularView<Eigen::Upper>();
+  eigenToCSC(triangular_sm, P_row_indices_, P_column_pointers_, P_csc_data_);
```
Levi-Armstrong: Does this change the behavior. Before the whole matrix was passed and now only the upper triangle is?

haudren:
Sorry for the late reply:
- I usually install OSQP directly from source as it's a simple C package. It is not packaged as debs, although they have binary releases ( https://bintray.com/bstellato/generic/OSQP/0.6.0#files/OSQP/0.6.0 ). Given that the dependencies are very light, I don't think it would be any issue to build on 16.04 / 18.04.
- For the upper triangle thing, no it does not change the behavior in my limited testing. I believe they were already only using the upper triangle, but now there is a check in place that rejects full matrices. In my experience with QP solvers, we used to only fill the upper triangle anyways as it's supposed to be symmetric, and then fill the other triangle if the solver requested it (usually for dense solvers).

Levi-Armstrong:
@haudren It looks like it fails CI because clang formatting needs to be ran. Once this is addressed it is ready to be merged.

haudren:
Formatting should be fixed now!

mpowelson:
@Levi-Armstrong I've tested this on my system and it seems to work fine on the examples

Levi-Armstrong:
@haudren Thank you for the contribution. We plan to make this the default optimizer because of the license.

@mpowelson Thank you for testing this. We will need to update existing system to the latest version of OSQP.

haudren:
@Levi-Armstrong : Note that after more thorough testing, I found some problems on which OSQP fails. Not sure if BPMPD would have been able to handle them, but I think if OSQP becomes the default, it would be nice to properly dump the problem when that happens. Debugging solver failures is difficult at the moment as only Gurobi properly fills `fail.lp`.

Unfortunately, there doesn't seem to be a function for that in OSQP, but I guess it can be implemented. If I get some time to work on it I'll try to submit a PR

arocchi:
In my experience, there are definitely cases in which trajopt succeeds with BPMPD/Gurobi but not with OSQP or qpOASES. Never had the time to try ipOpt to see where the success rate stands with it.

PR #152, opened 2019-12-27
Enable continuous collision checking for moving to moving objects

Levi-Armstrong:
This enables avoiding moving objects to moving objects when continuous collision checking is enabled.

### DIFF
--- a/trajopt/src/collision_terms.cpp
+++ b/trajopt/src/collision_terms.cpp
@@ -35,7 +35,7 @@ void DebugPrintInfo(const tesseract_collision::ContactResult& res,
   {
     std::printf("\n");
     std::printf("DistanceResult| %30s | %30s | %6s | %6s, %6s, %6s | %6s, %6s, "
-                "%6s | %6s, %6s, %6s | %7s | %6s, %6s, %6s | %6s, %6s, %6s |",
+                "%6s | %6s, %6s, %6s | %6s, %6s, %6s | %6s, %6s, %6s | %10s %10s |",
                 "LINK A",
                 "LINK B",
                 "DIST",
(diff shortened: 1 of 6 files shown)


trajopt/src/collision_terms.cpp
```diff
@@ -402,35 +427,107 @@ CastCollisionEvaluator::CastCollisionEvaluator(tesseract_kinematics::ForwardKine
...
-  tesseract_environment::EnvState::Ptr state0 = env_->getState(manip_->getJointNames(), sco::getVec(x, m_vars0));
-  tesseract_environment::EnvState::Ptr state1 = env_->getState(manip_->getJointNames(), sco::getVec(x, m_vars1));
-  for (const auto& link_name : adjacency_map_->getActiveLinkNames())
-    contact_manager_->setCollisionObjectsTransform(
-        link_name, state0->transforms[link_name], state1->transforms[link_name]);
+  Eigen::VectorXd s0 = sco::getVec(x, m_vars0);
+  Eigen::VectorXd s1 = sco::getVec(x, m_vars1);
+  tesseract_collision::ContactResultMap contact_results;
+  double dist = (s1 - s0).norm();
+  if (dist > longest_valid_segment_length_)
```
mpowelson: A comment stating what each of these blocks of code do would be nice.

mpowelson:
@Levi-Armstrong This looks nice. I'm excited to see how this performs in the wild.

Levi-Armstrong:
I added documentation to the new code.

Levi-Armstrong:
@mpowelson Two of the tests in trajopt were failing and after further investigations I found that it was bad to only keep the works case when checking intermediate state. The optimizer need all of the information to be able to correctly adjust the start and end state to avoid collision. The last commit correct this and now all tests pass.

PR #153, opened 2020-01-01
Remove the additional 0.04 added to contact distance threshold

Levi-Armstrong:
Reading through the old repository this was carried over because Bullet added this margin to the contact shapes. This makes since in the case of solving for the exact point of contact but is not required just for contact checking so removing. This should speed up the optimization for chases where you are always in close proximity.


Levi-Armstrong:
Yea, It needs to know when it gets close to an object so it can then start running one of the contact solvers. Some of the solvers will not work if the object is already in collision so I believe they add this buffer to insure it catches it just prior to collision to prevent this.

PR #154, opened 2020-01-09
Update casted collision evaluator to handle fixed start and end states

Levi-Armstrong:
This updates casted collision evaluator to handle fixed start or end states. Before we just would exclude but that is not ideal because you would periodically get post check failures  between the start and first waypoint and next to last waypoint and the end waypoint. 

Also this PR includes updating the constructors to pass eigen types by reference. The documentation states not to pass eigen types by value and I know in the past we have had eigen issues.

### DIFF
--- a/trajopt/include/trajopt/collision_terms.hpp
+++ b/trajopt/include/trajopt/collision_terms.hpp
@@ -8,6 +8,13 @@
 
 namespace trajopt
 {
+enum class CollisionEvaluatorType
+{
+  START_FREE_END_FREE = 0,  /**< @brief Both start and end state variables are free to be adjusted */
+  START_FREE_END_FIXED = 1, /**< @brief Only start state variables are free to be adjusted */
+  START_FIXED_END_FREE = 2  /**< @brief Only end state variables are free to be adjusted */
+};
+
 struct CollisionEvaluator
 {
   EIGEN_MAKE_ALIGNED_OPERATOR_NEW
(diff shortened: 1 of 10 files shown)


trajopt/src/collision_terms.cpp
```diff
@@ -528,8 +876,43 @@ void CastCollisionEvaluator::CalcCollisions(const DblVec& x, tesseract_collision
...
+                  {
+                    return false;
+                  }
+                  case CollisionExpressionEvaluatorType::START_FIXED_END_FREE:
+                  {
+                    if (r.cc_type[0] == tesseract_collision::ContinuousCollisionType::CCType_Time0)
+                      return true;
+
+                    if (r.cc_type[1] == tesseract_collision::ContinuousCollisionType::CCType_Time0)
+                      return true;
```
mpowelson: Should CCTYPE_NONE be removed?

mpowelson: Also, do these enums follow the correct style?

trajopt/src/collision_terms.cpp
```diff
@@ -421,6 +454,295 @@ void SingleTimestepCollisionEvaluator::Plot(const tesseract_visualization::Visua
...
+  double dist = (s1 - s0).norm();
+  long cnt = 2;
+  if (dist > longest_valid_segment_length_)
+  {
+    // Calculate the number state to interpolate
+    cnt = static_cast<long>(std::ceil(dist / longest_valid_segment_length_)) + 1;
+  }
+
+  // Get active link names
+  const std::vector<std::string>& active_links = adjacency_map_->getActiveLinkNames();
```
mpowelson: This is not in the CastCollisionEvaluator. Should it be added around 812?

trajopt/src/collision_terms.cpp
```diff
@@ -421,6 +454,295 @@ void SingleTimestepCollisionEvaluator::Plot(const tesseract_visualization::Visua
...
+        // If the contact pair does not exist in contact_results add it
+        if (p == contact_results.end() && !pair.second.empty())
+        {
+          contact_results[pair.first] = pair.second;
+        }
+        else
+        {
+          // Note: Must include all contacts through out the trajectory so the optimizer has all the information
+          //      to understand how to adjust the start and end state to move it out of collision. Originally tried
+          //      keeping the works case only but ran into edge cases where this does not work in the units tests.
```
mpowelson: typos
`works case` -> `worst case`

`through out` -> `throughout`

trajopt/src/collision_terms.cpp
Levi-Armstrong: Are you suggesting removing it from the enum class?

mpowelson: No, should it return true if it is that type.

Levi-Armstrong: Right now CCType_None is to represent objects that are static in the world. You should never have a case where you checking two shapes that are of CCType_None which is why I only check for the CCType_Time0 or CCType_Time1 depending on the condition.

trajopt/src/collision_terms.cpp
Levi-Armstrong: This has been restructured.

Levi-Armstrong:
@mpowelson I clean things up and added documentation to the collision evaluators. I also tested this on the examples and everything works as expected.

mpowelson:
Remember to remove CI commit

trajopt/include/trajopt/collision_terms.hpp
```diff
@@ -54,11 +100,57 @@ struct CollisionEvaluator
...
+
+  /**
+   * @brief Calculate the distance expressions when the end is free but the start is fixed
+   * @param x The current values
+   * @param exprs The returned expression
+   */
+  void CalcDistExpressionsEndFree(const DblVec& x, sco::AffExprVector& exprs);
+
+  /**
+   * @brief Calculate the distance expressions when the start and end is free
```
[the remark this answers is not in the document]
Levi-Armstrong: fixed

Levi-Armstrong:
I made the requested changes and remove the commit.

PR #155, opened 2020-01-09
Change Default Solver

mpowelson:
Gurobi is still the preferred solver, but the default solver assuming no other setup has been done on the computer will now be OSQP. This has been added to Trajopt_ext. If OSQP is not found at compile time, it will clone it into the build directory and compile it. 

The motivation behind this is that BPMPD has a noncommercial license. By changing the default solver to OSQP we limit the chance that someone uses BPMPD in a commercial application inadvertently. It will now require an intentional change.

The new order for AUTO_SOLVER is 
* Gurobi
* OSQP
* BPMPD
* qpOASES

### DIFF
--- a/trajopt_ext/osqp/CMakeLists.txt
+++ b/trajopt_ext/osqp/CMakeLists.txt
@@ -1,0 +1,23 @@
+cmake_minimum_required(VERSION 3.5.0)
+project(osqp)
+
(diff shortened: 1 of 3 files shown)


Levi-Armstrong:
I will wait on CI to pass before merging.

PR #156, opened 2020-01-15
Add trajopt_sco depend on osqp

mpowelson:


Levi-Armstrong:
@mpowelson It looks like moving GIT_SHALLOW worked. You could add a check or just remove it, I will leave that up to you. 

```
if(${CMAKE_VERSION} VERSION_LESS "3.6.0")
endif()
```

mpowelson:
Bummer. It's probably not worth it for osqp but it might be for a big package like Bullet with a long history.

Levi-Armstrong:
Lets just remove it.

mpowelson:
So now trajopt_sco is failing clang-tidy. I'll have to clean that up in a bit.

trajopt_sco/src/osqp_interface.cpp
```diff
@@ -124,31 +124,34 @@ void OSQPModel::updateConstraints()
...
-  sm.conservativeResize(m + n, Eigen::NoChange_t(n));
+  sm.conservativeResize(m_int + n_int, Eigen::NoChange_t(n));
 
   l_.clear();
   l_.resize(m + n, -OSQP_INFINITY);
   u_.clear();
   u_.resize(m + n, OSQP_INFINITY);
 
-  for (int i_cnt = 0; i_cnt < m; ++i_cnt)
+  for (int i_cnt = 0; i_cnt < m_int; ++i_cnt)
```
Levi-Armstrong: could you use size_t as the type to avoid cast inside the for loop?

mpowelson: I changed it. There are still static casts, but now there are 2 to int instead of 3 to size_t

mpowelson:
@Levi-Armstrong Do you know what version of clang-tidy is installed on the CI docker?

trajopt_sco/src/osqp_interface.cpp
Levi-Armstrong: Oh, did not change much.

Levi-Armstrong:
Not sure. What ever the default Debian version is for the os.

Levi-Armstrong:
Oh, The CI is set to version 8. If you are on kinetic you will need to modify the run_clang format to use clang version 8. In the future it would be good to modify the script so it checks the version and print a error if you are using the wrong version.

mpowelson:
@Levi-Armstrong See what you think. If you don't like that last commit I can get rid of it. If you do, I can squash it with the NOLINT commit.

Levi-Armstrong:
Lets add the last commit in a separate PR.

mpowelson:
See #157

PR #160, opened 2020-01-23
Add a safety margin buffer to collision evaluators

schornakj:
Adds a `safety_margin_buffer` parameter to the Trajopt collision evaluators. This is an additional buffer that is added to the safety margin when checking if contacts should be evaluated as collisions. Contact distances within `safety_margin + safety_margin_buffer` will be evaluated by the optimization but will not contribute a cost. The purpose of this change is to help the optimization stay out of collision when the safety margin distance is very small. The default value is set to 0.05m.

Other changes:
- Fix a missing OSQP dependency in `trajopt_sco`
- Fix a bug where the optimization would fail in cases when it actually converged with all constraints satisfied.

### DIFF
--- a/trajopt/include/trajopt/problem_description.hpp
+++ b/trajopt/include/trajopt/problem_description.hpp
@@ -568,6 +568,11 @@ struct CollisionTermInfo : public TermInfo
    */
   double longest_valid_segment_length = 0.5;
 
+  /** @brief A buffer added to the collision margin distance. Contact results that are within the safety margin buffer distance but greater than the
+  safety margin distance (i.e. close but not in collision) will be evaluated but will not contribute costs to the optimization problem.
+  This helps keep the solution away from collision constraint conditions when the safety margin distance is small.*/
+  double safety_margin_buffer = 0.05;
+
   /** @brief Set the contact test type that should be used. */
   tesseract_collision::ContactTestType contact_test_type = tesseract_collision::ContactTestType::ALL;
 
--- a/trajopt/include/trajopt/problem_description.hpp
+++ b/trajopt/include/trajopt/problem_description.hpp
@@ -568,9 +568,10 @@ struct CollisionTermInfo : public TermInfo
    */
   double longest_valid_segment_length = 0.5;
 
-  /** @brief A buffer added to the collision margin distance. Contact results that are within the safety margin buffer distance but greater than the
-  safety margin distance (i.e. close but not in collision) will be evaluated but will not contribute costs to the optimization problem.
-  This helps keep the solution away from collision constraint conditions when the safety margin distance is small.*/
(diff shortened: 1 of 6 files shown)


schornakj:
@Levi-Armstrong Looks like CI is failing because of a rosdep-install error. Do you know what might be wrong?

trajopt_sco/package.xml
```diff
@@ -9,6 +9,7 @@
   <depend>osqp</depend>
   <depend>trajopt_utils</depend>
   <depend>libjsoncpp-dev</depend>
+  <depend>osqp</depend>
```
Levi-Armstrong: osqp is already added at line 9. This may be the issues with ci.

schornakj: Whoops, good catch!

schornakj:
@Levi-Armstrong Is this good to go?

trajopt/include/trajopt/problem_description.hpp
```diff
@@ -568,6 +568,12 @@ struct CollisionTermInfo : public TermInfo
    */
   double longest_valid_segment_length = 0.5;
 
+  /** @brief A buffer added to the collision margin distance. Contact results that are within the safety margin buffer
+  distance but greater than the safety margin distance (i.e. close but not in collision) will be evaluated but will not
+  contribute costs to the optimization problem. This helps keep the solution away from collision constraint conditions
+  when the safety margin distance is small.*/
+  double safety_margin_buffer = 0.05;
+
```
Levi-Armstrong: This only applies to constraints not costs correct?

trajopt/src/problem_description.cpp
```diff
@@ -1782,7 +1787,8 @@ void CollisionTermInfo::hatch(TrajOptProb& prob)
                                                        prob.GetVarRow(i, 0, n_dof),
                                                        prob.GetVarRow(i + 1, 0, n_dof),
                                                        expression_evaluator_type,
-                                                       discrete_continuous);
+                                                       discrete_continuous,
+                                                       safety_margin_buffer);
```
Levi-Armstrong: For costs should we eliminate this parameter  and just pass zero to the Collision evaluator?

schornakj: I made the default value of this parameter zero in the Tesseract CollisionCostConfig in [this PR](https://github.com/ros-industrial-consortium/tesseract/pull/215). There might be some reason to have a non-zero buffer distance for collision costs, so I don't know that we should disable it entirely.

trajopt/include/trajopt/problem_description.hpp
schornakj: Right now it applies to both. I'm not sure what a good default value is, so 0.05m is kind of arbitrary.

PR #161, opened 2020-01-24
Add LP format printing to OSQP solver

haudren:
I have had a few solving errors with OSQP but could not easily debug. Using this PR allows to print LP information when solving fails.

I created a few helpers in osqp_printing and use those when solving fails. Unfortunately the default stream operators do not have the LP formatting in mind (e.g. no operator between coefficient and variables), I think this PR is not OSQP-specific and could be refactored / extended if you think it is worth it.

Another point that is more general: LP format is meant for ease of readability, not for reproducibility. I think we could write to a more widely-supported like MPS / AMPL / ... but it requires more work as the format are not as straightforward.


Levi-Armstrong:
@haudren What are your thought on adding the relavent code to a stream operator for the AffExp and QuadExp. Also as you mentioned it looks like this could be used by all solvers, if that is the case if we add the stream operators to the AffExp and QuadExp then we could add your print function the base class. What do you think?

haudren:
Hum I looked into that, but there is already a stream operator for both AffExpr and QuadExpr and I did not want to change the current behavior.

For sure what I wrote can be used in the base class, and overriden by Gurobi's own `writeModel`.

Levi-Armstrong:
>Hum I looked into that, but there is already a stream operator for both AffExpr and QuadExpr and I did not want to change the current behavior.

I am good with replacing those with what you have created.

haudren:
So I tried to do the above, no problems to actually replace the individual calls to printing AffExpr and QuadExpr. Unfortunately, the printing of constraints will need to be solver specific, as we cannot access the list of constraints from the abstract `Model`, but only from the derived classes... I'll try to wrap this up tonight.

haudren:
@Levi-Armstrong : I just updated this PR with version 2 ! I now use the stream operators, but as mentioned above, the implementation for OSQP had to be specialized to retrieve the constraints correctly. This now seems to output a reasonable LP format that will hopefully be helpful to OSQP users.

Note that I also marked `writeToFile` as `const`, it helped me spot a typo in my implementation which could have had some pretty gnarly consequences ^^

trajopt_sco/src/solver_interface.cpp
```diff
@@ -132,21 +132,58 @@ std::ostream& operator<<(std::ostream& o, const Cnt& c)
 
 std::ostream& operator<<(std::ostream& o, const AffExpr& e)
 {
-  o << e.constant;
+  std::string sep = "";
+  if (e.constant != 0)
```
Levi-Armstrong: Should [boost](https://www.boost.org/doc/libs/1_63_0/libs/math/doc/html/math_toolkit/float_comparison.html) be used for comparing floats?

Levi-Armstrong: Everything else looks good.

haudren: Hum I know float comparison is a tricky field, but in this case I think we should print exactly what is sent to the solver, without making assumptions as to what will be considered zero / non-zero. Using relative float comparison might for example hide issues with uninitialized values.

Levi-Armstrong: Make sense in this case

PR #164, opened 2020-01-30
Fix issue in osqp interface to update osqp_data object with new A and P

Levi-Armstrong:


schornakj:
Something's still wrong, I get a runtime error [here](https://github.com/ros-industrial-consortium/trajopt_ros/blob/3c27463d36693f82c5946fc215eb55977499db10/trajopt_sco/src/osqp_interface.cpp#L98) when resetting `P_`.

Levi-Armstrong:
@schornakj This should fix the issue.

PR #165, opened 2020-02-04
Add ability to use weighted sum jacobian calculation for contact link pairs

Levi-Armstrong:
The current implementation for the collision term added an equation for each contact results. This means for a given link pair it could have multiple contacts if the links contain multiple collision objections. This results in a large number of equation to solve for some environments. This enables the ability to use the weighted sum for a given link pair so you only end up with one equation per link pair significantly reducing the number of equations to solve.

### DIFF
--- a/trajopt/include/trajopt/problem_description.hpp
+++ b/trajopt/include/trajopt/problem_description.hpp
@@ -557,6 +557,12 @@ struct CollisionTermInfo : public TermInfo
   /** @brief Indicate the type of collision checking that should be used. */
   CollisionEvaluatorType evaluator_type;
 
+  /**
+   * @brief Use the weighted sum for each link pair. This reduces the number equations added to the problem
+   * When enable it is good to start with a coefficient of 1 otherwise 20 is a good starting point.
+   */
+  bool use_weighted_sum = false;
+
   /** @brief Indicated if a step is fixed and its variables cannot be changed */
   std::vector<int> fixed_steps;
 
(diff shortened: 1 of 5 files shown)


trajopt/src/collision_terms.cpp
```diff
@@ -265,6 +267,358 @@ void CollisionsToDistanceExpressions(sco::AffExprVector& exprs,
...
+  for (const auto& pair : dist_results)
+  {
+    double worst_dist{ std::numeric_limits<double>::max() };
+    double total_weight_a{ 0 }, total_weight_b{ 0 };
+    bool found_a{ false }, found_b{ false };
+
+    Eigen::VectorXd dist_grad_a = Eigen::VectorXd::Zero(manip_->numJoints());
+    Eigen::VectorXd dist_grad_b = Eigen::VectorXd::Zero(manip_->numJoints());
+
+    const Eigen::Vector2d& data = safety_margin_data_->getPairSafetyMarginData(pair.first.first, pair.first.second);
```
mpowelson: A more descriptive variable name might be useful here. I had to look back to figure out what data was.

trajopt/include/trajopt/problem_description.hpp
```diff
@@ -557,6 +557,9 @@ struct CollisionTermInfo : public TermInfo
   /** @brief Indicate the type of collision checking that should be used. */
   CollisionEvaluatorType evaluator_type;
 
+  /** @brief Use the weighted sum for each link pair. This reduces the number equations added to the problem */
+  bool use_weighted_sum = true;
```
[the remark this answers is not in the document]
Levi-Armstrong: Sure.

trajopt/src/collision_terms.cpp
Levi-Armstrong: I added comments everywhere this is created.

Levi-Armstrong:
@mpowelson I changed the default settings and add comments.

PR #169, opened 2020-02-13
Check init_info.data.size() when using JOINT_INTERPOLATED

mpowelson:
Allow either 1 x DOF or DOF x 1. Previously it would segfault if you gave it a TrajArray corresponding to one joint position (a single row).


mpowelson:
@Levi-Armstrong I don't think CI failing is related to this PR. It looks like cast cost isn't moving the link.

mpowelson:
@Levi-Armstrong It's passing now. You must have restarted it.

PR #170, opened 2020-02-18
Use std::move when calling addLink

mpowelson:
The unit tests were broken in a recent Tesseract PR.


gavanderhoorn:
@Levi-Armstrong: would it be an idea to include `trajopt_ros` in the CI build of `tesseract`? That way these kinds of breakages could be easily detected.

You could add a row to the allowed-to-fail builds. A breakage would then not prevent a merge, but would still be flagged.

mpowelson:
So I enabled a daily cron job in Travis for TrajOpt to help us catch this kind of thing. Tesseract already takes so long, I'd hate to add another one. I was noticing yesterday that if we have more than one open PR on Tesseract we exceed the max concurrent jobs, so we just have to wait for the first 1.5hr build to finish.

mpowelson:
Though practically, I'm not sure how long it actually takes to run the trajopt unit tests. It might be negligible since we are building it anyway.

PR #171, opened 2020-02-21
Changes for Gurobi compatibility with Ubuntu 18.04

skaruparthi:
vars is of type sco::SizeTVec which is std::vector of type "long unsigned int"
IntVec is of type sco::IntVec which is std::vector of type "int"

Auto conversion not working with Gurobi on Ubumtu 18.04

### DIFF
--- a/trajopt_sco/include/trajopt_sco/solver_interface.hpp
+++ b/trajopt_sco/include/trajopt_sco/solver_interface.hpp
@@ -234,7 +234,11 @@ Model::Ptr createModel(ModelType model_type = ModelType::AUTO_SOLVER);
 
 SizeTVec vars2inds(const VarVector& vars);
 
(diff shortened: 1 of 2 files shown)


mpowelson:
It looks like Clang-tidy failed. You'll probably want to build locally with these flags on to see what the problems are.
-DTRAJOPT_ENABLE_CLANG_TIDY=ON -DTRAJOPT_ENABLE_TESTING=ON

schornakj:
I think you also have to rename the function in the places where the version that returns an `IntVec` is used.

Levi-Armstrong:
@schornakj Did you PR address these the changes?

schornakj:
@Levi-Armstrong @skaruparthi Yes, this is addressed in #174.

Levi-Armstrong:
addressed by #174

PR #173, opened 2020-03-06
Add GetGradient function to CollisionEvaluator

Levi-Armstrong:
This exposed the ability to get the gradient from the collision evaluator to support alternative use cases. It reduce code duplication within the evaluator.

### DIFF
--- a/trajopt/src/collision_terms.cpp
+++ b/trajopt/src/collision_terms.cpp
@@ -159,6 +159,144 @@ void DebugPrintInfo(const tesseract_collision::ContactResult& res,
   std::printf("\n");
 }
 
+GradientResults CollisionEvaluator::GetGradient(const Eigen::VectorXd& dofvals,
+                                                const tesseract_collision::ContactResult& contact_result,
+                                                const Eigen::Vector2d& data,
+                                                bool isTimestep1)
+{
+  GradientResults results(data);
+  for (std::size_t i = 0; i < 2; ++i)
+  {
+    tesseract_environment::AdjacencyMapPair::ConstPtr it = adjacency_map_->getLinkMapping(contact_result.link_names[i]);
+    if (it != nullptr)
+    {
+      results.gradients[i].has_gradient = true;
+
+      Eigen::MatrixXd jac;
+      jac.resize(6, manip_->numJoints());
+
+      // Calculate Jacobian
+      manip_->calcJacobian(jac, dofvals, it->link_name);
+
+      // Need to change the base and ref point of the jacobian.
+      // When changing ref point you must provide a vector from the current ref
+      // point to the new ref point.
+      results.gradients[i].scale = 1;
+      Eigen::Isometry3d link_transform = contact_result.transform[i];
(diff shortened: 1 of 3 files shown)


trajopt/src/collision_terms.cpp
```diff
@@ -159,6 +159,225 @@ void DebugPrintInfo(const tesseract_collision::ContactResult& res,
...
+    // temp2 = link_transform * itB->transform.inverse();
+    // assert(temp1.isApprox(temp2, 0.0001));
+
+    // Eigen::MatrixXd jac_test;
+    // jac_test.resize(6, manip->numJoints());
+    // tesseract_kinematics::numericalJacobian(jac_test, world_to_base, *manip, dofvals, itB->link_name,
+    // res.nearest_points_local[1]); bool check = jac.isApprox(jac_test, 1e-3); assert(check == true);
+
+    results.gradients[1].gradient = contact_result.normal.transpose() * jac.topRows(3);
+  }
```
marip8: Consider consolidating this into a for loop since it looks like you're doing the same operations for the first and second collision link

trajopt/src/collision_terms.cpp
```diff
@@ -159,6 +159,225 @@ void DebugPrintInfo(const tesseract_collision::ContactResult& res,
...
```
marip8: Same comment as above about consolidating into a for loop. It also looks like you're duplicating a lot of code from previous functions. Is there a way to reformat this overload to utilize the logic the "main" function rather than copying the content?

trajopt/src/collision_terms.cpp
```diff
@@ -159,6 +159,225 @@ void DebugPrintInfo(const tesseract_collision::ContactResult& res,
...
+    // Eigen::Isometry3d test_link_transform, temp1, temp2;
+    // manip->calcFwdKin(test_link_transform, dofvals, itA->link_name);
+    // temp1 = world_to_base * test_link_transform;
+    // temp2 = link_transform * itA->transform.inverse();
+    // assert(temp1.isApprox(temp2, 0.0001));
+
+    // Eigen::MatrixXd jac_test;
+    // jac_test.resize(6, manip->numJoints());
+    // tesseract_kinematics::numericalJacobian(jac_test, world_to_base, *manip, dofvals, itA->link_name,
+    // res.nearest_points_local[0]); bool check = jac.isApprox(jac_test, 1e-3); assert(check == true);
```
marip8: Consider removing commented code if it is no longer useful

trajopt/src/collision_terms.cpp
Levi-Armstrong: Done

trajopt/src/collision_terms.cpp
Levi-Armstrong: I added the for loop.

trajopt/src/collision_terms.cpp
Levi-Armstrong: I use it periodically as a sanity check when I make changes to this function.

Levi-Armstrong:
@marip8 I made the for loop change.

Issue #168, opened 2020-02-11
Return type of function doesn't match the variable type assigned while calling the function


skaruparthi:
In "trajopt_sco/src/gurobi_interface.cpp"
IntVec inds = vars2indsI(vars);

vars is of type sco::SizeTVec which is std::vector of type "long unsigned int"
IntVec is of type sco::IntVec which is std::vector of type "int"

This is causing the build to fail when using Gurobi.
Trying to install on Ubuntu 18.04 with Gurobi 9.0 and Latest version of Trajopt

Levi-Armstrong:
Sorry, I do not have a license to Gurobi to confirm. If you created a PR to fix the issue we will get it merged into master.

skaruparthi:
I have made a couple of changes and got it working. How do I go about submitting a pull request? I tried the normal way but it just said permission denied. Sorry if it is a stupid question. Thanks in advance

Levi-Armstrong:
I would fork this repository and push your changes to your fork of this repository. Once that is done you then can create a PR from your fork to this repository.

Levi-Armstrong:
@skaruparthi Did #174 resolve your issue?

PR #174, opened 2020-03-12
Fixes for Gurobi 9.0.1

schornakj:
- Update `FindGUROBI.cmake` to find Gurobi 9.0 and to more generally match the example Gurobi provides [here](https://support.gurobi.com/hc/en-us/articles/360039499751-CMake-C-C-compilation-of-Gurobi-projects).
- Change the `vars2inds` and `cnts2inds` functions so they returns void and writes the output vector of indices to a parameter. Added new versions of these function that output a `SizeTVec`, which complements the existing functions that outputs a `IntVec`. This is a solution to the return type issue described in #168.
- Modify usages of these functions to reflect new return type and parameters.
- Add `static_cast`s where warnings were complaining about implicit conversions between types changing values.

### DIFF
--- a/trajopt_sco/CMakeLists.txt
+++ b/trajopt_sco/CMakeLists.txt
@@ -66,7 +66,7 @@ add_library(${PROJECT_NAME} SHARED ${SCO_SOURCE_FILES})
 
 if (GUROBI_FOUND)
   target_link_libraries(${PROJECT_NAME} PRIVATE ${GUROBI_LIBRARIES})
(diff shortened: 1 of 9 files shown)


schornakj:
As a side note, we may want to consider setting up a way to consistently build and test support for Gurobi.

Levi-Armstrong:
@schornakj Clang-Tidy is not happy

schornakj:
> @schornakj Clang-Tidy is not happy

Looks like I was able to satisfy it

PR #178, opened 2020-03-26
Add CalcCollisions that takes only joint values instead of Vars

mpowelson:
This adds CalcCollision functions to the TrajOpt collision evaluators that take only Eigen types. This allows them to be used outside of the current Trajopt setup (and independently of Vars). Note that the constructor still requires that a VarVector be passed in (it can be empty if using the evaluator this way). We may want to revisit this with a more comprehensive refactor if we ever pursue something like IFOPT to break the dependency on sco.

### DIFF
--- a/trajopt/include/trajopt/collision_terms.hpp
+++ b/trajopt/include/trajopt/collision_terms.hpp
@@ -389,6 +389,12 @@ public:
                            sco::AffExprVector& exprs,
                            AlignedVector<Eigen::Vector2d>& exprs_data) override;
   void CalcCollisions(const DblVec& x, tesseract_collision::ContactResultMap& dist_results) override;
+  /**
+   * @brief Given joint names and values calculate the collision results for this evaluator
+   * @param dof_vals Joint values set prior to collision checking
+   * @param dist_results Contact Results Map
+   */
+  void CalcCollisions(const Eigen::Ref<Eigen::VectorXd>& dof_vals, tesseract_collision::ContactResultMap& dist_results);
   void Plot(const tesseract_visualization::Visualization::Ptr& plotter, const DblVec& x) override;
   sco::VarVector GetVars() override { return vars0_; }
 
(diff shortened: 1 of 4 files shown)


trajopt/include/trajopt/collision_terms.hpp
```diff
@@ -389,6 +389,12 @@ struct SingleTimestepCollisionEvaluator : public CollisionEvaluator
                            sco::AffExprVector& exprs,
                            AlignedVector<Eigen::Vector2d>& exprs_data) override;
   void CalcCollisions(const DblVec& x, tesseract_collision::ContactResultMap& dist_results) override;
+  /**
+   * @brief Given joint names and values calculate the collision results for this evaluator
+   * @param dof_vals Joint values set prior to collision checking
+   * @param dist_results Contact Results Map
+   */
+  void CalcCollisions(const Eigen::Ref<Eigen::VectorXd> dof_vals, tesseract_collision::ContactResultMap& dist_results);
```
Levi-Armstrong: should be `const Eigen::Ref<Eigen::VectorXd>&`

trajopt/include/trajopt/collision_terms.hpp
```diff
@@ -419,6 +425,15 @@ struct CastCollisionEvaluator : public CollisionEvaluator
...
   void CalcCollisions(const DblVec& x, tesseract_collision::ContactResultMap& dist_results) override;
+  /**
+   * @brief Given joint names and values calculate the collision results for this evaluator
+   * @param dof_vals0 Joint values for state0
+   * @param dof_vals1 Joint values for state1
+   * @param dist_results Contact Results Map
+   */
+  void CalcCollisions(const Eigen::Ref<Eigen::VectorXd> dof_vals0,
+                      const Eigen::Ref<Eigen::VectorXd> dof_vals1,
+                      tesseract_collision::ContactResultMap& dist_results);
```
Levi-Armstrong: should be `const Eigen::Ref<Eigen::VectorXd>&`

trajopt/include/trajopt/collision_terms.hpp
```diff
@@ -449,6 +464,15 @@ struct DiscreteCollisionEvaluator : public CollisionEvaluator
...
```
Levi-Armstrong: should be `const Eigen::Ref<Eigen::VectorXd>&`

PR #182, opened 2020-05-27
Whitelist trajopt_sqp and trajopt_ifopt

mpowelson:


mpowelson:
@Levi-Armstrong This looks like it fixed it.

PR #194, opened 2020-07-07
github-ci: melodic and noetic initial commit

Minipada:
Hi,
Happy to participate in the World ROS I day activity with Brisa Robotics.

Here are my proposed changes to build on Melodic and Noetic.
I know melodic was not requested but it was a good starting point for me to get into Github actions.
In the ci file, you will notice melodic uses upstream but Noetic uses a tesseract fork I created (see https://github.com/brisa-robotics/tesseract/tree/noetic)

Some infos:
The reason for that is the switch to python3 in Noetic which makes packages.xml non compatible anymore.
python becomes python3
python-numpy becomes python3-numpy
orocos_kdl becomes liborocos-kdl-dev

For python3-numpy, see:
https://github.com/ros/rosdistro/pull/24035

Regarding orocos, see mentions here:
https://github.com/ros/rosdistro/pull/24009
https://github.com/ros/geometry2/pull/447
and example there
https://github.com/ros/geometry2/pull/447/files

Finally,
I ignored `ifopt` in Noetic at the moment, it has no release since lunar (source: https://github.com/ethz-adrl/ifopt-release). You probably want to release one or we need to add an entry in rosdistro at least if lunar version is used

Correct me if I'm wrong, but now there are 2 possibilities on how to proceed.
1. Do you create a noetic branch on tesseract where I'd do a PR, and then change the github action file to point there?
2. Or do you make all python2 deprecated and i do a PR on master where python3 is mandatory
Last point: should i remove the travis file?

Since this is part of World ROS I day, I'm available on google meet and irc if you want to discuss into more details.

Build result: https://github.com/brisa-robotics/trajopt_ros/runs/846092749?check_suite_focus=true


.github/workflows/main.yml
```diff
@@ -0,0 +1,62 @@
...
+    strategy:
+      fail-fast: false
+      matrix:
+        env:
+          - {CI_NAME: Bionic-Build,
+              OS_NAME: ubuntu,
+              OS_CODE_NAME: bionic,
+              ROS_DISTRO: melodic,
+              ROS_REPO: main,
+              UPSTREAM_WORKSPACE: "github:ros-industrial-consortium/tesseract#master github:ros-industrial-consortium/trajopt_ros#master
```
mpowelson: We should be able to do something like this to avoid building two versions of Trajopt.
```
             UPSTREAM_WORKSPACE: 'github:swri-robotics/descartes_light#master github:Jmeyer1292/opw_kinematics#master github:ethz-adrl/ifopt#master',
             TARGET_WORKSPACE: '. github:ros-industrial-consortium/tesseract#master',
```

.github/workflows/main.yml
```diff
@@ -0,0 +1,62 @@
...
+              PARALLEL_TESTS: false,
+              CCACHE_DIR: "/home/runner/work/trajopt_ros/trajopt_ros/Bionic-Build/.ccache",
+              UPSTREAM_CMAKE_ARGS: "-DCMAKE_BUILD_TYPE=Release",
+              TARGET_CMAKE_ARGS: "-DCMAKE_BUILD_TYPE=Debug"}
+          - {CI_NAME: Focal-Build,
+              OS_NAME: ubuntu,
+              OS_CODE_NAME: focal,
+              ROS_DISTRO: noetic,
+              ROS_REPO: main,
+              UPSTREAM_WORKSPACE: "github:brisa-robotics/tesseract#noetic github:ros-industrial-consortium/trajopt_ros#master
```
mpowelson: To address the package.xml differences, we may be able to use something like. I've never used and/or in a package.xml but it [looks](https://www.ros.org/reps/rep-0149.html) like it should work
```
  <depend condition="$ROS_VERSION == 1 and $ROS_DISTRO != noetic">orocos_kdl</depend>
  <depend condition="$ROS_VERSION == 2 or  $ROS_DISTRO == noetic">liborocos-kdl-dev</depend>
```

and 
```
  <depend condition="$ROS_PYTHON_VERSION == 2">python</depend>
  <depend condition="$ROS_PYTHON_VERSION == 3">python3</depend>
```

@Levi-Armstrong may have a cleaner solution though.

.github/workflows/main.yml
Minipada: Yep thanks, I understood it a bit after, it's done in https://github.com/ros-industrial-consortium/trajopt_ros/pull/194/commits/85ee28ea02e17f981e721b8651c9ff8225b2a2a4

mpowelson: Thanks. I didn't see your commits until I submitted this.

Minipada: It seems it works for both. Build result updated at https://github.com/brisa-robotics/trajopt_ros/runs/846092749?check_suite_focus=true

.github/workflows/main.yml
Minipada: I didn't know about this, that's great. I will patch the branch as soon as Levi answers.

Levi-Armstrong: I have created a [PR](https://github.com/ros-industrial-consortium/tesseract/pull/319) against Tesseract adding the conditionals but it appear to break industrial_ci for some reason. I have created an [issue](https://github.com/ros-industrial/industrial_ci/issues/546) on industrial_ci though not sure why this would have caused an issue.

mpowelson: [Tesseract #319](https://github.com/ros-industrial-consortium/tesseract/pull/319) should have fixed this. You should be able to switch it back to upstream master.

Minipada: This seems to work. Thanks.

Minipada:
To complete the ticket, Foxy was also requested. I could make it work but I had to apply changes to the Dockerfile. Should I make a PR on the dockerfile repo first?

mpowelson:
> To complete the ticket, Foxy was also requested. I could make it work but I had to apply changes to the Dockerfile. Should I make a PR on the dockerfile repo first?

Yeah if you don't mind, that would be great.

Minipada:
I just noticed I commented on another PR.
Please ping me when foxy is on https://hub.docker.com/r/lharmstrong/tesseract/tags, i'll add the github action CI build right after.

PR #195, opened 2020-07-09
Add cmake_common_scripts to rosinstall

mpowelson:
This is a new dependency of opw_kinematics


Levi-Armstrong:
It looks like iwyu needs to be added to skip keys.

PR #197, opened 2020-07-15
New and improved refactored testing framework with simple nlp problems

rkliman:
Used feedback to create a better version of the code. Osqp_eigen has some trouble with the basic test so it is not run. This is likely due to some problem setup that is different than what osqp_eigen is used to.


mpowelson:
Make sure you are using clang-format-8. If you run the script in the repo it should work.

PR #204, opened 2020-10-02
Catch exception when OSQP setup fails

mpowelson:


mpowelson:
While this does catch the exception, I still get double free errors immediately after. I'll have to find those

mpowelson:
It appears that if osqp_setup returns false, then `osqp_workspace_ != nullptr` and `osqp_cleanup(osqp_workspace_);` will cause a double free error. Annoyingly, I can't seem to make it happen when osqp is built in debug.

Levi-Armstrong:
What about in the try catch if an exception is thrown set `osqp_workspace_ = nullptr`?

mpowelson:
> What about in the try catch if an exception is thrown set `osqp_workspace_ = nullptr`?

That's what I did. It does solve the problem, but I wasnt sure if that was a memory leak depending on what in osqp_workspace got set.

Levi-Armstrong:
Yea probably need to look at what that function does, most likely they are not checking if something is null before trying to delete it.

Levi-Armstrong:
I looked at the osqp_cleanup and did not see anything obvious. It looks like they are checking everything before deleting it on the master branch.

Levi-Armstrong:
@mpowelson Can you rebase on the latest?

mpowelson:
Sure. Should I just add `osqp_workspace_ = nullptr` before we throw?

mpowelson:
Otherwise, this doesn't really fix anything. It still crashes just instead of a missed exception it is a double free

Levi-Armstrong:
@mpowelson Sorry, I forget to push changes updating the new Github Action to use the correct branches for feature/CommandLanguage. If you rebase now it should build.

mpowelson:
@Levi-Armstrong So setting it to nullptr apparently doesn't always work. The error is a `OSQP_DATA_VALIDATION_ERROR`

Levi-Armstrong:
What was the setup error you were receiving?

Levi-Armstrong:
It looks like no memory is allocated if either of the errors occur

```
OSQP_DATA_VALIDATION_ERROR
OSQP_SETTINGS_VALIDATION_ERROR

```

Levi-Armstrong:
I think we should check the return type [here](https://github.com/ros-industrial-consortium/trajopt_ros/blob/e6bb2807d0ba16f367c1642b69becee679135e15/trajopt_sco/src/osqp_interface.cpp#L193-L197) and if it is either of the types above we explicitly set `osqp_workspace_=nullptr`;

mpowelson:
Is that the problem? Are we trying to free it when it was never allocated?

Unfortunately, I don't have an example to recreate this. The piece of code I had that was doing this was basically just doing IK. Maybe 1-2% of the time this would pop up.

Levi-Armstrong:
I think so.

Levi-Armstrong (approved):

mpowelson:
Added. I will say that I apparently thought that setting it to nullptr wouldn't always fix it. But maybe there were multiple failure conditions and I just needed to do it on these 2.

PR #207, opened 2020-10-05
Add building and running of tests in github actions

Levi-Armstrong:


Levi-Armstrong:
@mpowelson It looks like the interface to IFOPT changed [here](https://github.com/ethz-adrl/ifopt/blob/8976a52669bd2a269254ef86ddc1e522a93fccac/ifopt_core/include/ifopt/problem.h#L175). You must not pass a bool indicating if finite difference should be used. Do you know if this should be true or false for where we are using it?

Levi-Armstrong:
@mpowelson It was fixed upstream.

mpowelson:
Yeah, they added that in. We actually already had our own utility to do that, but I'd have to compare the two to see if they are interchangeable. We use ours in `CartPosConstraint::CalcJacobianBlock` for example. So it looks like there are 3 ways to do it now
1) Our utility
2) Their utility
3) Solver specific settings- e.g. `ipopt.SetOption("jacobian_approximation", "finite-difference-values");`

Levi-Armstrong:
It looks like it will be more work to get CI to pass on Focal due to new checks in clang-tidy. I will have a new machine soon that will make it easy to address the required changes but will leave it as is as long as Bionic passes.

PR #210, opened 2020-10-19
Add windows support and CI build

Levi-Armstrong:
This updates everything to build on windows along with switching to cmake_common_scripts. I am not able to build all tests because need to figure out how to get ipopt install on windows, but will save that for another PR once it is added to tesseract_ext.


colin-lewis-19:
Works with the current feature/CommandLanguage branches

PR #212, opened 2020-10-20
Improve const-correctness of reference passing.

a-price:
This fix allows const-correctness in downstream projects, for constructions like
```
const Eigen::VectorXd home_position = ...
const auto initial_states = trajopt::interpolate(home_position, ...
```

One potential concern would be non-contiguous memory silently converting to a copy rather than failing to compile. From https://eigen.tuxfamily.org/dox/classEigen_1_1Ref.html: "In the const case, if the input expression does not match the above requirement, then it is evaluated into a temporary before being passed to the function." I didn't see anywhere that would be a concern, and the copy might be faster in some contexts due to improved locality/vectorization, but it's something to be aware of.

### DIFF
--- a/trajopt/include/trajopt/collision_terms.hpp
+++ b/trajopt/include/trajopt/collision_terms.hpp
@@ -404,7 +404,7 @@ public:
    * @param dof_vals Joint values set prior to collision checking
    * @param dist_results Contact Results Map
    */
-  void CalcCollisions(const Eigen::Ref<Eigen::VectorXd>& dof_vals, tesseract_collision::ContactResultMap& dist_results);
+  void CalcCollisions(const Eigen::Ref<const Eigen::VectorXd>& dof_vals, tesseract_collision::ContactResultMap& dist_results);
   void Plot(const tesseract_visualization::Visualization::Ptr& plotter, const DblVec& x) override;
   sco::VarVector GetVars() override { return vars0_; }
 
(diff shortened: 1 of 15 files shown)


Levi-Armstrong:
Looks good, I will need to port these change to the feature/CommandLanguage branch.

a-price:
By the way, I think there's some issues with compile flags leaking out as interface flags or applying to 3rd party deps; that's what's causing the build errors with newer GCC versions. `VHACD` in particular had complaints about `memcpy`, some private, undefined but not deleted constructors, and a cast constness issue. That could probably be resolved in a couple of ways:
* Fix `trajopt_macros.cmake` to not expose `-Wall -Wextra` stuff to its dependents. (This looks like it shouldn't happen; could be an issue with my CMake.)
* Modify `TRAJOPT_IGNORE_WARNINGS_PUSH` to ignore new GCC warnings
* Minimal fix for the current complaints in `VHACD`. (I went with this approach on my machine.)
* More general cleanup of `VHACD`: run `clang-tidy`, port their custom small-vector optimized class to e.g. [absl::InlinedVector](https://github.com/abseil/abseil-cpp/blob/master/absl/container/inlined_vector.h), etc.

Levi-Armstrong:
There are, I just fixed this in my latest PR to TrajOpt and I am working on this for Tesseract branch feature/CommandLanguage. They are currently marked public and I am breaking up these into public/private variables so the warning/error do not get passed to those that include this package.

a-price:
Awesome, thanks! By the way, this is some very nice work y'all have done.

Levi-Armstrong:
Thank you, All recent improvements have been being pushed to the feature/CommandLanguage branch for both tesseract, trajopt, and descartes if you interested in trying out the new motion planning framework. We are very close to having it wrapped up and ready to merge into master.

PR #213, opened 2020-10-21
Add public compiler option -mno-avx

Levi-Armstrong:
Add back the disabling avx instructions


marrts (approved):
This fixed my Eigen alignment issues

PR #216, opened 2020-10-28
Improve const-correctness of reference passing see PR #212

Levi-Armstrong:


trajopt_ifopt/include/trajopt_ifopt/constraints/collision_constraint.h
```diff
@@ -73,7 +73,7 @@ class CollisionConstraintIfopt : public ifopt::ConstraintSet
    * @brief Fills the jacobian block associated with the constraint
    * @param jac_block Block of the overall jacobian associated with these constraints
    */
-  void CalcJacobianBlock(const Eigen::Ref<Eigen::VectorXd>& joint_vals, Jacobian& jac_block) const;
+  void CalcJacobianBlock(const Eigen::Ref<const Eigen::VectorXd>& joint_vals, Jacobian& jac_block) const;
```
mpowelson, line 76: Slightly unrelated, but should this be `Eigen::Ref<Jacobian>` ? [This](https://stackoverflow.com/questions/21132538/correct-usage-of-the-eigenref-class) answer indicates that for writable reference, that is preferred.

mpowelson (approved):

trajopt_ifopt/include/trajopt_ifopt/constraints/collision_constraint.h
Levi-Armstrong: Yep if it is an Eigen Type. I will make the change.

Levi-Armstrong: Well, it looks like we cannot use the Ref because we are resizing the matrix size.

mpowelson: That makes me feel better. I wondered why I didn't do that.

PR #219, opened 2020-11-24
Trajopt_ifopt: Add Collision Evaluators

mpowelson:
This moves the collision evaluators from trajopt to trajopt_ifopt. Currently only the Discrete one has been done (ie, the single timestamp one). It is likely that this code can/should be significantly refactored at some point. However, the goal of this PR was more to define the TrajOptCollisionConfig to be used in the tesseract planner.


mpowelson:
Requires [Tesseract #456](https://github.com/ros-industrial-consortium/tesseract/pull/456)

Levi-Armstrong:
@mpowelson Is this ready to review? I was not sure since CI is not passing.

mpowelson:
I thought it was but like you said, CI isn't passing. It looks like it has the same issue Colin's PR has. If you'd like to take a look and/or finish it up that would be appreciated. Otherwise it will probably be after the new year before I get back to it.

Levi-Armstrong:
Sure, Once finished with the examples in tesseract_ros I can possibly take a look.

Levi-Armstrong:
@mpowelson I believe if you rebase the issue should be resolved with the unit tests.

trajopt_ifopt/include/trajopt_ifopt/constraints/collision_evaluators.h
```diff
@@ -0,0 +1,459 @@
...
+   * @param dist_results Contact Results Map
+   */
+  void CalcCollisions(const Eigen::Ref<const Eigen::VectorXd>& dof_vals,
+                      tesseract_collision::ContactResultMap& dist_results);
+
+private:
+  tesseract_collision::DiscreteContactManager::Ptr contact_manager_;
+};
+
+///**
```
Levi-Armstrong, line 373: Should the comment code below be removed?

trajopt_ifopt/src/collision_evaluators.cpp
```diff
@@ -0,0 +1,1304 @@
...
+    const Eigen::Vector2d data = { dist, coeff };
+    auto end = std::remove_if(
+        pair.second.begin(), pair.second.end(), [&data, this](const tesseract_collision::ContactResult& r) {
+          return (!((data[0] + collision_config_.collision_margin_buffer) > r.distance));
+        });
+    pair.second.erase(end, pair.second.end());
+  }
+}
+
+//////////////////////////////////////////
```
Levi-Armstrong, line 597: Should the commented code below be removed?

Levi-Armstrong (commented):
looks good. A few minor questions. Is this now separated from the old code?

mpowelson:
I'm not sure if it is or not. We are still using some utilties it looks like. Those won't be too hard to remove though.

The commented code are evaluators that still need to be ported. trajopt_ifopt currently is only set up from Discrete. So LVSContinuous and LVSDiscrete still need to be done.

Levi-Armstrong:
not sure why windows build is not happy. I assume either missing a header or need to set cxx version to 17 maybe.

Levi-Armstrong:
It looks like it does not like something in the trajopt_sqp types.h possible the enum class at the bottom.

trajopt_ifopt/src/collision_evaluators.cpp
Levi-Armstrong: The commented code is what is left to be implemented and will be kept until then.

trajopt_ifopt/include/trajopt_ifopt/constraints/collision_evaluators.h
Levi-Armstrong: The commented code is what is left to be implemented and will be kept until then.

mpowelson:
> It looks like it does not like something in the trajopt_sqp types.h possible the enum class at the bottom.

So what do you recommend I do? 

Looking at [C2059](https://docs.microsoft.com/en-us/cpp/error-messages/compiler-errors-1/compiler-error-c2059?view=msvc-160), maybe Windows has a define for one of these? 
```
  RUNNING,
  TRUST_REGION_CONVERGED,
  NLP_CONVERGED,
  ITERATION_LIMIT,
  ERROR,
  CALLBACK_STOPPED
```

mpowelson:
For convenience, here is the error
https://github.com/ros-industrial-consortium/trajopt_ros/pull/219/checks?check_run_id=1623604924#step:3:10227

gavanderhoorn:
> Looking at [C2059](https://docs.microsoft.com/en-us/cpp/error-messages/compiler-errors-1/compiler-error-c2059?view=msvc-160), maybe Windows has a define for one of these?

I'm almost certain it's `ERROR`.

mpowelson:
That seems likely but also rather unfortunate.  I guess I can rename it to OPTIMIZATION_ERROR or something

Levi-Armstrong:
> > It looks like it does not like something in the trajopt_sqp types.h possible the enum class at the bottom.
> 
> So what do you recommend I do?
> 
> Looking at [C2059](https://docs.microsoft.com/en-us/cpp/error-messages/compiler-errors-1/compiler-error-c2059?view=msvc-160), maybe Windows has a define for one of these?
> 
> ```
>   RUNNING,
>   TRUST_REGION_CONVERGED,
>   NLP_CONVERGED,
>   ITERATION_LIMIT,
>   ERROR,
>   CALLBACK_STOPPED
> ```

Maybe try defining a type for the enum class as int. If that does not work maybe prefix all or just error with `SQP_`?

mpowelson:
Would [this](https://stackoverflow.com/questions/1543736/how-do-i-temporarily-disable-a-macro-expansion-in-c-c) solve the problem? And if so, is that a really bad idea? SQPStatus::ERROR won't have this problem right?

mpowelson:
I'm thinking just renaming it is probably best and cleanest

gavanderhoorn:
> Would [this](https://stackoverflow.com/questions/1543736/how-do-i-temporarily-disable-a-macro-expansion-in-c-c) solve the problem? And if so, is that a really bad idea?

is a bit hackish, yes.

> I'm thinking just renaming it is probably best and cleanest

Agreed. "ERROR" is rather ambiguous in any case.

Error what?

mpowelson:
Turns out a couple of those weren't even being used. I cleaned those up a bit and pushed them in a new commit. The new set is 

```
  RUNNING,         /**< Optimization is currently running */
  NLP_CONVERGED,   /**< NLP Successfully converged */
  ITERATION_LIMIT, /**< SQP Optimization reached iteration limit */
  QP_SOLVER_ERROR, /**< QP Solver failed */
  CALLBACK_STOPPED /**< Optimization stopped because callback returned false */
```

mpowelson:
Same problem in QPSolverStatus. I just pushed that change as well.

mpowelson:
@Levi-Armstrong Looks like trajopt_sco isn't building tests?

mpowelson:
@Levi-Armstrong That did not fix it. trajopt_sco tests are still not being built.

Levi-Armstrong:
I would remove trajopt_sco from the CI run tests for now.

Levi-Armstrong:
@mpowelson If you squash the last two commits together this should be good to merge.

PR #226, opened 2020-12-31
Rename `dofs_fixed` parameter

marip8:
This PR changes the name of the `dofs_fixed` parameter to `fixed_timesteps` to more accurately represent its function. This variable is used [here](https://github.com/ros-industrial-consortium/trajopt_ros/blob/b84e4ae30fa658d2911f3e375bb6009dd8ba7084/trajopt/src/problem_description.cpp#L496-L507) constrain all of the optimization variables in each specified time step to their initial values.

This PR also updates the name of the avoid singularity cost that is printed to the terminal during optimization

### DIFF
--- a/trajopt/include/trajopt/collision_terms.hpp
+++ b/trajopt/include/trajopt/collision_terms.hpp
@@ -404,7 +404,8 @@ public:
    * @param dof_vals Joint values set prior to collision checking
    * @param dist_results Contact Results Map
    */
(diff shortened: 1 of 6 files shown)


marip8:
@Levi-Armstrong there might be a circular dependency with this PR: it looks like `tesseract_python` fails to build because this PR changes the name of a variable in the `BasicInfo` class. My guess is that the [corresponding `tesseract` PR](https://github.com/ros-industrial-consortium/tesseract/pull/502) will also fail because it will try to reference the newly named parameter. What's the best way forward with this?

Levi-Armstrong (approved):

PR #227, opened 2021-01-01
Rename dofs_fixed parameter

Levi-Armstrong:
Port #226 changes to feature/CommandLanguage branch


Levi-Armstrong:
This will not pass until the one in tesseract is also merged.

PR #228, opened 2021-01-02
Improve fixed timesteps and dofs

Levi-Armstrong:


Levi-Armstrong:
@marip8 I found an issue in the fixed_timesteps. Also added  the ability to fix a joint for all timesteps which I think is what the original implementation was doing. Need to back port this master.

PR #245, opened 2021-02-24
Implement max_time

mpowelson:


trajopt_sco/include/trajopt_sco/optimizers.hpp
```diff
@@ -101,7 +102,8 @@ struct BasicTrustRegionSQPParameters
   double max_merit_coeff_increases;
 
   double merit_coeff_increase_ratio;  // ratio that we increate coeff each time
-  double max_time;                    // not yet implemented
+  /** @brief Max time in seconds that the optimizer will run*/
+  double max_time;
```
Levi-Armstrong, line 106: Should this be initialized to `std::numeric_limits<double>::max()`?

Levi-Armstrong (commented):
Looks good, just one comment. I did not see where the max_time was being initialized.

trajopt_sco/include/trajopt_sco/optimizers.hpp
mpowelson: Odd. Yeah it should be.

mpowelson: Oh that is already being set in `BasicTrustRegionSQPParameters::BasicTrustRegionSQPParameters()`. That's why I didn't do it in this commit.   `max_time = static_cast<double>(INFINITY);`

Levi-Armstrong: Ah ok.

Levi-Armstrong (approved):

PR #246, opened 2021-02-24
Feature/line constraint

colin-lewis-19:
This adds a constraint defined by two two points that calculates the target poses' distance to the line between them.


colin-lewis-19:
@mpowelson Can you give this a look?

trajopt_ifopt/include/trajopt_ifopt/constraints/cartesian_line_constraint.h
```diff
@@ -0,0 +1,198 @@
+﻿/**
+ * @file cartesian_position_constraint.h
+ * @brief The cartesian position constraint
```
mpowelson, line 3: Filename

trajopt_ifopt/include/trajopt_ifopt/constraints/cartesian_line_constraint.h
```diff
@@ -0,0 +1,198 @@
...
+   * @param var_set Name of the var_set to which the jac_block is associated
+   * @param jac_block Block of the overal jacobian associated with these constraints and the var_set variable
+   */
+  void FillJacobianBlock(std::string var_set, Jacobian& jac_block) const override;
+
+  void SetLine(const Eigen::Isometry3d& Point_A, const Eigen::Isometry3d& Point_B);
+
+  void SetLinePose(const Eigen::Isometry3d& line_point);
+
+  void SetTargetPose(const Eigen::Isometry3d& target_pose);
```
mpowelson, line 139: Doxygen.

Should this be SetLinePoint?

trajopt_ifopt/src/cartesian_line_constraint.cpp
```diff
@@ -0,0 +1,222 @@
+/**
+ * @file cartesian_position_constraint.h
+ * @brief The cartesian position constraint
```
mpowelson, line 3: Same

trajopt_ifopt/src/cartesian_line_constraint.cpp
```diff
@@ -0,0 +1,222 @@
...
+  else
+  {
+    line_point = point_a_.translation() + mag * line_norm;
+  }
+
+  // pose error is the vector from the new_pose to nearest point on line AB, line_point
+  Eigen::Vector3d pose_err = (line_point - new_pose.translation()).array().abs();
+  // @TODO: Handle orientation
+  Eigen::Vector3d rot_ = Eigen::Vector3d(0.0, 0.0, 0.0);
+  Eigen::VectorXd err = concat(pose_err, rot_);
```
mpowelson, line 95: Might put this todo in the header doxygen too

trajopt_ifopt/src/cartesian_line_constraint.cpp
```diff
@@ -0,0 +1,222 @@
...
+    // the partial derivative of the error function. Note that the rotational portion is the only part
+    // that is required to be modified per the paper.
+    Eigen::Isometry3d pose_err = line_point_inv_ * tf0;
+    Eigen::Vector3d rot_err = calcRotationalError(pose_err.rotation());
+    for (int c = 0; c < jac0.cols(); ++c)
+    {
+      auto new_pose_err = addTwist(pose_err, jac0.col(c), 1e-5);
+      Eigen::VectorXd new_rot_err = calcRotationalError(new_pose_err.rotation());
+      jac0.col(c).tail(3) = ((new_rot_err - rot_err) / 1e-5);
+    }
```
mpowelson, line 179: If you don't have rotation in the error you should comment it in the jacobian as well.

trajopt_ifopt/src/cartesian_line_constraint.cpp
```diff
@@ -0,0 +1,222 @@
...
+  {
+    line_point = point_b_.translation();
+  }
+  else if (mag < 0)
+  {
+    line_point = point_a_.translation();
+  }
+  else
+  {
+    line_point = point_a_.translation() + mag * line_norm;
```
mpowelson, line 88: Should line_point_ be stored here?

trajopt_ifopt/src/cartesian_line_constraint.cpp
```diff
@@ -0,0 +1,222 @@
...
+Eigen::VectorXd CartLineConstraint::GetValues() const
+{
+  Eigen::VectorXd joint_vals = this->GetVariables()->GetComponent(position_var_->GetName())->GetValues();
+  return CalcValues(joint_vals);
+}
+
+void CartLineConstraint::SetLinePose(const Eigen::Isometry3d& line_point)
+{
+  line_point_ = line_point;
+  line_point_inv_ = line_point_.inverse();
```
mpowelson, line 108: Why would you want to set a line_point_ other than the one calculated above?

mpowelson (changes requested):

trajopt_ifopt/src/cartesian_line_constraint.cpp
colin-lewis-19: It should. However, this class extends the ifopt::ConstraintSet. The CalcValues function that is inherited from this object is a const method. To preserve this, I could not save off the line_point here as I intended.

marip8:
This looks like an useful constraint. A few comments:

1. The unit test really needs to have an optimization that solves for a trajectory with this constraint and proves that all points in the trajectory actually lie on the line. The current unit tests seem to only be checking that the error calculation works and that the Jacobian is filled out
2. Not sure if this is in scope, but adding orientation to the constraint would not be difficult if you use quaternion SLERP. You know the closest point from the current TCP location to the desired line, so you also know the fraction of the distance from the start point along that line. You can SLERP between the start and end orientations to that distance fraction and compare the interpolated orientation to the current TCP orientation.
3. Most of the function naming in this class does not follow the ROS or C++ style guidelines (mostly capitalized function names rather than camelCase). In the interest of uniformity, I would suggest changing those

Levi-Armstrong:
> The unit test really needs to have an optimization that solves for a trajectory with this constraint and proves that all points in the trajectory actually lie on the line. The current unit tests seem to only be checking that the error calculation works and that the Jacobian is filled out

@colin-lewis-19 You should be able to use one of the existing test to accomplish this.

> Not sure if this is in scope, but adding orientation to the constraint would not be difficult if you use quaternion SLERP. You know the closest point from the current TCP location to the desired line, so you also know the fraction of the distance from the start point along that line. You can SLERP between the start and end orientations to that distance fraction and compare the interpolated orientation to the current TCP orientation.

    If you're not going to work with orientation here, I would suggest working in terms of vectors rather than poses (i.e. point_a, point_b, etc. are Eigen::Vector3d). It would probably make it less ambiguous about whether or not orientation is being considered

I also think this is important feature to have for this constraint.

>     3\. Most of the function naming in this class does not follow the ROS or C++ style guidelines (mostly capitalized function names rather than camelCase). In the interest of uniformity, I would suggest changing those

Just to provide some history. Early on we looked into this because trajopt_ifopt builds on the IFOPT library which unfortunately uses a different style than ROS. To avoid mixing styles it was decided use IFOPT's for this particular package and at the time we found language in the style guide that this is acceptable when building on another interface which does not us the ROS style. @mpowelson may be able to provide more detail.

mpowelson:
2) At one point I thought @colin-lewis-19 had a version that did orientation. I guess I assumed he removed it because it wasn't working with the intention that we would add it back eventually. I'm sure he would accept pull requests.

1) Obviously integrations tests would be great. However, as long as the getValue and getJacobians are correct, I would be happy.

3) Yeah we are inheriting from an IFOPT type, so we can't do much about the style guide.

trajopt_ifopt/src/cartesian_line_constraint.cpp
colin-lewis-19: This was included for thoroughness. I don't see many use cases beyond unit tests currently

colin-lewis-19:
I would be interested in pursuing SLERP but it may be some time before I can address it. If that is not a requirement, I think this may be ready.

colin-lewis-19:
I believe I have addressed the concerns. It now handles orientation & has a corresponding unit test.

Levi-Armstrong:
@colin-lewis-19 Can you rebase on master and then we can track down why the unit tests are not passing?

colin-lewis-19:
No, I was tying to debug clang tidy.

PR #247, opened 2021-03-05
Remove checkJoints check no longer needed

Levi-Armstrong:
A while back we updated kinematics so that forward and inverse would still calculated the results even if the joint values are outside the joint limits for the purpose of numerically calculated the jacobian. This check is no longer needed.


mpowelson (approved):
If you say so.

PR #248, opened 2021-03-18
Update for changes with CollisionMarginData

Levi-Armstrong:


trajopt_utils/package.xml
```diff
@@ -13,6 +13,7 @@
   <build_depend>ros_industrial_cmake_boilerplate</build_depend>
   <build_depend>eigen</build_depend>
   <build_export_depend>eigen</build_export_depend>
+  <depend>libboost-program-options</depend>
```
mpowelson, line 16: Does this need the -dev version?

mpowelson (approved):

trajopt_utils/package.xml
Levi-Armstrong: Good catch. I updated it to the following but going to wait for windows CI to finish because it is failing for some reason related to osqp with an odd error.

``` xml
  <build_depend>libboost-program-options-dev</build_depend>
  <exec_depend>libboost-program-options</exec_depend>
```

Levi-Armstrong:
I will merge if the windows build fails again. It look like something must have change upstream in ros-win and will need more time to look into what changed.

PR #252, opened 2021-04-05
Fix JointPosIneqConstraint

### LINKED ISSUE #251: JointPosIneqConstraint does not correctly use lower_limits
I fixed the cost a while back but the constraint is also wrong.

mpowelson:
Closes #251. I'm pretty sure this fixes it. However OSQP was still having a hard time solving QPs with this in it. I decided to not worry about it since I hope to upgrade to trajopt_ifopt before too long anyway.

### DIFF
--- a/.github/workflows/windows_noetic_build.yml
+++ b/.github/workflows/windows_noetic_build.yml
@@ -37,8 +37,8 @@ jobs:
 
         rosdep install -q --from-paths . --ignore-src -y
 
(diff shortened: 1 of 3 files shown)


Levi-Armstrong (approved):

mpowelson:
@Levi-Armstrong I didn't realize that this is broken on master. I'm guessing the updates to the IK interface did it. We probably need to add a check to it to take the closest solution.

Levi-Armstrong:
@mpowelson Were you planning to fix the unit tests as part of this PR?

Levi-Armstrong:
This is good to merge if CI passes.

Issue #193, opened 2020-07-06
Add other collision evaluators to trajopt_ifopt


mpowelson:

mpowelson:
@steviedale is tackling this one.

mpowelson:
@Levi-Armstrong @colin-lewis-19 Has any progress been made on this one? Ideally we would also update this to use the tesseract_collision::CollisionCheckConfig before [Tesseract #443](https://github.com/ros-industrial-consortium/tesseract/pull/443)

Levi-Armstrong:
I don't think so. I took a look a week ago to see what it would take but I did not start.

mpowelson:
The final thing to do here is to allow the constraint to take any of these. Currently only discrete is allowed

Issue #183, opened 2020-05-27
Trajopt_ifopt TODO list


mpowelson:
- [x] Collision - Add other collision evaluators (See #193)
- [ ] Collision Callback - Collision is the only one that is missing (See #192)
- [x] Print debug info - trajopt_sco prints a bunch of debugging information that developers have become accustomed to. Some of this has been added to trajopt_sqp, but some like the cost associated with individual convex costs, will require some work to implement
- [ ] [Add Hessian to IFOPT](https://github.com/ethz-adrl/ifopt/issues/41)
- [x] Look into adding caching to functions like Problem::EvaluateCostFunction
- [x] Add coeff support to constraints. 
  - [x] Collision Constraint
  - [x] Cartesian Position Constraint
  - [x] Inverse Kinematics Constraint
  - [x] Joint Position Constraint
  - [x] Joint Velocity Constraint
  - [x] Joint Acceleration Constraint
  - [x] Joint Jerk Constraint
- [x] Joint Acceleration ( #275 )
- [x] Joint Jerk (#275 )

Levi-Armstrong:
@mpowelson Is this what is remaining for  completing trajopt_ifopt where it has the same functionality as trajopt?

mpowelson:
Big things I can think of are #192 and #193. Smaller issues, we are missing the dynamic cartesian constraint as well as joint acceleration. There are probably other things that will come up where you can semantically do the same thing but it looks a little different in the optimization (e.g. the old trajopt start_fixed flag), but I think the other collision evaluators are really the main thing that is missing before it will work for the majority of cases.

Of course time parameterization is also something that will need to be reworked.

colin-lewis-19:
I'm going to start chipping away at this, starting with #192

mpowelson:
Other possible performance improvements.

- [ ] Test out other linear solvers. OSQP supports [MKL Parsido].(https://osqp.org/docs/interfaces/linear_systems_solvers.html#linear-system-solvers-setting). "MKL Pardiso is an efficient multi-threaded linear system solver that works well for large scale problems part of the Intel Math Kernel Library"
- [ ] IFOPT problems don't change size, so we could use [OSQP code generation](https://osqp.org/docs/codegen/index.html) to generate a problem once and solve it repeatedly. I'm not sure if this is supposed to be faster or is just for deploying the QP solver in an external codebase

simonschmeisser:
[...]

PR #256, opened 2021-04-13
Shrink trust region if convex solver fails

mpowelson:
I'm not sure if this is a good idea or not. I'm testing it out, and it seems promising. However, I'm putting this here so others can test it out.

### DIFF
--- a/trajopt_sco/include/trajopt_sco/optimizers.hpp
+++ b/trajopt_sco/include/trajopt_sco/optimizers.hpp
@@ -100,6 +100,8 @@ struct BasicTrustRegionSQPParameters
   // constraint violation is less than this, we're done
   /** @brief Max number of times that the constraints' cost will be increased */
   double max_merit_coeff_increases;
+  /** @brief Max number of times the QP solver can fail before optimization is aborted*/
+  int max_qp_solver_failures;
 
   double merit_coeff_increase_ratio;  // ratio that we increate coeff each time
   /** @brief Max time in seconds that the optimizer will run*/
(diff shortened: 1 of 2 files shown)


Levi-Armstrong:
This looks reasonable. I have been seeing more QP solver failures since we switched to OSQP so I will test. Also if this works I think the default trust region may be to large.

mpowelson:
For my problem it seems like this works about half the time. About half the time of it fails once it will fail a bunch of times in a row. They other half it will fail once or twice but then succeed with the smaller trust region. 

I'm curious what other things solvers like snopt do to mitigate this. Maybe we could relax constraints, solve, and then reapply them? I'd really rather not put too much more effort into this solver though. If we do anything major is rather do it in trajopt_sqp.

Levi-Armstrong:
It may be worth trying where we start out with everything as a cost for a few iteration and then witch them to constraints. I believe this what you are referring to by relax the constraints? 

> I'd really rather not put too much more effort into this solver though. If we do anything major is rather do it in trajopt_sqp.
I agree, do we have a list of what is left to do before making the jump?

Levi-Armstrong:
@gavanderhoorn You recent asked this question in the annual meeting. Do you have thoughts on the subject?

mpowelson:
For me, the limitation is time to update my code. I think trajopt_ifopt is probably useable as is. We do have #183 where we are keeping a list. I think #193 is probably the main thing you would miss.

mpowelson:
I've also pondered seeing if we should put trajopt_sqp in its own repo and push back an interface to it to ifopt like they have for ipopt and snopt. The free SQP optimizer might be useful to people outside of trajopt, and we might get input from actual optimization experts.

mpowelson:
Removing the wip since I am using this, and it seems to work fine. I can keep using it for a week or so and see if anything pops up if you want, or we can merge it.

Levi-Armstrong (approved):
If you have been using it without issues then I am good with you merging it.

PR #257, opened 2021-04-20
Fix trajopt_ifopt inverse kinematics constraint unit test

Levi-Armstrong:
The original unit test was setting the seed outside the joint limits.


mpowelson (approved):
Thanks for fixing this

PR #259, opened 2021-04-22
Joint Acceleration Constraint

ben-greenberg:
Took a stab at adding this acceleration constraint, to smooth trajectories instead of using the velocity constraint (which was causing solutions to enter an 'unfeasible' state). I basically just modified the joint_velocity_constraint to make this. Seems like I can use it, though I'm not sure if its doing what I want. Happy to get any feedback on this.


mpowelson:
So adding variables for velocity is something that we have talked about. However this is probably not what you are wanting. Since all of the other constraints operate on positions, you won't have any way of tying the goal waypoints to the velocity variables. Trajopt may optimize it, but you'll just end up with velocities that don't correspond to the position variables.

What you probably want is to implement acceleration constraints using central finite difference like we do in the old trajopt version. You can check out the old cost in trajectory_costs.cpp. I think the math should be pretty much the same minus the squaring of the results.
```
DblVec JointAccEqConstraint::value(const DblVec& xvec)
{
  // Convert vector from optimization to trajectory
  Eigen::MatrixXd traj = getTraj(xvec, vars_);
  // Takes diff b/n the subsequent rows and subtract each row by the targets_ vector
  Eigen::MatrixXd diff =
      (diffAxis0(diffAxis0(traj.block(first_step_, 0, last_step_ - first_step_ + 1, traj.cols())))).rowwise() -
      targets_.transpose();
  // Squares it, multiplies it by a diagonal matrix of coefficients, and converts to vector
  return util::toDblVec((diff.array().square()).matrix() * coeffs_.asDiagonal());
}
```

mpowelson:
> Thanks for the guidance! A few questions as I've been trying to get up to speed on what was done for the velocity constraint.
> 
> 1. `joint_velocity_constraint.h` defines the velocity as `Joint velocity is calculated as v = th_1 - th_0`

This is just saying that the velocity is the position at t1 - position at t0. However, the dt is assumed to be 1.

>    Where did this equation come from, and is there a similar one for acceleration?
> 2. I'm also unclear on what the `single_step` calculated in `joint_velocity_constraint.cpp` is representing. I see it is defined as `Eigen::VectorXd single_step = vals1 - vals2;`? I would have thought that the step would be `vals2 - vals1` if it is representing the difference between joint positions at a certain point, but that change seems to break things when running.

I'm not sure why this is that way. It might be a bug.

> 3. If we use the central finite distance as you suggested, I think we'll need to change add an additional step for the forward and backward differences in position. Does that sound like the correct approach?

You should be able to use the math I posted above (or [here](https://en.wikipedia.org/wiki/Finite_difference) for example). Calculating velocity is baked into the equation.

trajopt_ifopt/src/joint_acceleration_constraint.cpp
```diff
@@ -0,0 +1,111 @@
...
+TRAJOPT_IGNORE_WARNINGS_PUSH
+#include <console_bridge/console.h>
+TRAJOPT_IGNORE_WARNINGS_POP
+
+namespace trajopt
+{
+JointAccelConstraint::JointAccelConstraint(const Eigen::VectorXd& targets,
+                                           std::vector<JointPosition::ConstPtr> position_vars,
+                                           const std::string& name)
+  : ifopt::ConstraintSet(static_cast<int>(targets.size()) * static_cast<int>(position_vars.size() - 1), name)
```
mpowelson, line 37: I think this should be `position_vars.size - 2`?

mpowelson (commented):
This is looking really good. I'd ask that you add some basic unit tests to at least check the jacobian. We need to break this out into a generic test suite, but here is an example from the cartesian position test.

```
    // Calculate jacobian numerically
    auto error_calculator = [&](const Eigen::Ref<const Eigen::VectorXd>& x) { return constraint->CalcValues(x); };
    trajopt::Jacobian num_jac_block = trajopt::calcForwardNumJac(error_calculator, joint_position_mod, 1e-4);

    // Compare to constraint jacobian
    {
      trajopt::Jacobian jac_block(num_jac_block.rows(), num_jac_block.cols());
      constraint->CalcJacobianBlock(joint_position_mod, jac_block);
      EXPECT_TRUE(jac_block.isApprox(num_jac_block, 1e-3));
    }
```

trajopt_ifopt/src/joint_acceleration_constraint.cpp
ben-greenberg: I agree, it seems like that should be a `- 2`, but when I run with that the cost gets enormous, whereas when I run with a `- 1` the cost behaves as expected. I'm looking to see if there are more places aside from here and line 57 where this change needs to be propagated, but I haven't had luck finding it yet. Any thoughts why it seems to behave well with the `- 1`?

mpowelson: Not really. Maybe whatever problem you are testing it on has a high acceleration on the last point? You might try making a unit test that defines the variables to be some known function and then make sure that getValue returns the correct value. For instance, make position_vars follow a cubic and make sure the values of this constraint increase linearly

Levi-Armstrong:
@ben-greenberg Can you rebase on master?

Levi-Armstrong:
replaced by #275

PR #260, opened 2021-05-06
Fully integrate support for LVS collision constraints

Levi-Armstrong:


Levi-Armstrong:
@mpowelson I am getting closer  but still have two unit tests failing. The first is the numerical ik test which has the same results for several iterations but then they go in different directions so there may be a difference in the trust region solver which I may need a little help tracking down.

Levi-Armstrong:
Also, It took me a while to figure out how to use the `SquaredCost` with a constraint which requires you to call `LinkWithVariableSet` on the constraint before passing it to the `SquaredCost`. Is there a reason not to pass the problem into the constructor of the `SquaredCost` and let it call `LinkWithVariableSet`?

mpowelson:
I think that would be fine. I guess you have to have a problem in order to call getCost anyway right?

mpowelson:
> @mpowelson I am getting closer but still have two unit tests failing. The first is the numerical ik test which has the same results for several iterations but then they go in different directions so there may be a difference in the trust region solver which I may need a little help tracking down.

Let me know if I can help out.

Levi-Armstrong:
> I think that would be fine. I guess you have to have a problem in order to call getCost anyway right?

Yea, I don't think there is a case where you would not already have one. Another option is see if the maintainer would be willing to make the method LinkWithVarables virtual then we could overload it. Do you think it is worth proposing?

mpowelson:
> > I think that would be fine. I guess you have to have a problem in order to call getCost anyway right?
> 
> Yea, I don't think there is a case where you would not already have one. Another option is see if the maintainer would be willing to make the method LinkWithVarables virtual then we could overload it. Do you think it is worth proposing?

If we can fix it just by changing our interface, I would probably just do that to limit the amount of code we have to write. Whichever you think is cleaner is fine with me though.

Levi-Armstrong:
> > @mpowelson I am getting closer but still have two unit tests failing. The first is the numerical ik test which has the same results for several iterations but then they go in different directions so there may be a difference in the trust region solver which I may need a little help tracking down.
> 
> Let me know if I can help out.

So I have tracked down one difference. The `updateNLPVariableBounds` produces different bounds than the original OSQP interface. I am trying to track down how this is done in the original, but if you have time to take look also you may be more familiar with the code.

Levi-Armstrong:
Well, so it was not an issue with the updateNLPVariableBounds. The issue was I was not setting the bounds on the JointPositionVariable. After fixing the issue in the test I was able to go further comparing the results and found that the new `TrustRegionSQPSolver` is producing a different result when entering the trust-region loop code. The  OSQP Eigen library produces an error when we stepOptimization(nlp) is called a second time within this loop because it calls initSolver a second time after it has already been initialized. After this, it diverges. @mpowelson Do you know how we should address the issue? It is in trust_region_sqp_solver.cpp around line 100.

mpowelson:
> within

I seem to remember I had issues with this as well. I have a comment there about trying to remove the `qp_solver->clear()`. I guess you need to move all of the clear and all of the qp solver setup inside the loop. Or preferably, figure out why we have to clear the workspace each time and fix that.

trajopt_sco/src/osqp_interface.cpp
```diff
@@ -16,7 +16,7 @@ TRAJOPT_IGNORE_WARNINGS_POP
 namespace sco
 {
 const double OSQP_INFINITY = std::numeric_limits<double>::infinity();
-const bool SUPER_DEBUG_MODE = false;
+const bool SUPER_DEBUG_MODE = true;
```
mpowelson, line 19: Probably meant to turn this off

trajopt_sco/src/optimizers.cpp
```diff
@@ -17,7 +17,7 @@ TRAJOPT_IGNORE_WARNINGS_POP
 
 namespace sco
 {
-const bool SUPER_DEBUG_MODE = false;
+const bool SUPER_DEBUG_MODE = true;
```
mpowelson, line 20: Probably meant to turn this off

trajopt_optimizers/trajopt_sqp/test/test_suite_utils.hpp
```diff
@@ -0,0 +1,31 @@
+#include <trajopt_utils/macros.h>
```
mpowelson, line 1: I just realized we are missing licenses on a bunch of files in the new trajopt code. Looks like mostly in the unit tests

trajopt/test/numerical_ik_unit.cpp
```diff
@@ -0,0 +1,129 @@
...
+using namespace tesseract_environment;
+using namespace tesseract_collision;
+using namespace tesseract_kinematics;
+using namespace tesseract_visualization;
+using namespace tesseract_scene_graph;
+
+bool plotting = false; /**< Enable plotting */
+static const double LONGEST_VALID_SEGMENT_LENGTH = 0.05;
+
+class NumericalIKTest : public testing::TestWithParam<const char*>
```
mpowelson, line 36: Note to myself and posterity. This was moved from planning_unit.cpp

trajopt_ifopt/src/discrete_collision_constraint.cpp
```diff
@@ -0,0 +1,264 @@
...
+  {
+    case GradientCombineMethod::SUM:
+    {
+      for (tesseract_collision::ContactResult& dist_result : dist_results)
+      {
+        double dist = collision_evaluator_->GetCollisionConfig().collision_margin_data.getPairCollisionMargin(
+            dist_result.link_names[0], dist_result.link_names[1]);
+        double coeff = collision_evaluator_->GetCollisionConfig().collision_coeff_data.getPairCollisionCoeff(
+            dist_result.link_names[0], dist_result.link_names[1]);
+        err[0] += std::max<double>(((dist - dist_result.distance) * coeff), 0.);
```
mpowelson, line 99: I think the sign needs to be flipped here.

Also, I'm not sure the weighted sum makes sense. Penetration distances greater than the dist setting (currently negative) will still result in an error. I think we want to take the max before the square. Also, I would probably rename it to squared sum or something. 

|Dist | Dist result | error | std::max(error, 0) | std::max(error^2,0) | 
| --------------- | --------------- | --------------- | --------------- | --------------- | 
| 1 | 1.1 | -0.1 | 0 | 0.01|
| 1 | 0.5 |  0.5 | 0.5 | .25 |
| 1 | -0.5 | 1.5 | 1.5 | 2.25 | 
| 0 | 0.5 | -0.5 | 0 | .25 |
| 0 | -0.5 | 0.5 | 0.5 | .25 |

trajopt_ifopt/src/discrete_collision_constraint.cpp
```diff
@@ -0,0 +1,264 @@
...
+    {
+      for (tesseract_collision::ContactResult& dist_result : dist_results)
+      {
+        double dist = collision_evaluator_->GetCollisionConfig().collision_margin_data.getPairCollisionMargin(
+            dist_result.link_names[0], dist_result.link_names[1]);
+        double coeff = collision_evaluator_->GetCollisionConfig().collision_coeff_data.getPairCollisionCoeff(
+            dist_result.link_names[0], dist_result.link_names[1]);
+        double d = std::max<double>(((dist - dist_result.distance) * coeff), 0.);
+        if (d > err[0])
+          err[0] = d;
```
mpowelson, line 125: This seems to be a misnomer. It doesn't average the error. It actually takes the max.

trajopt_ifopt/src/discrete_collision_constraint.cpp
```diff
@@ -0,0 +1,264 @@
...
+        double dist = collision_evaluator_->GetCollisionConfig().collision_margin_data.getPairCollisionMargin(
+            dist_result.link_names[0], dist_result.link_names[1]);
+        double coeff = collision_evaluator_->GetCollisionConfig().collision_coeff_data.getPairCollisionCoeff(
+            dist_result.link_names[0], dist_result.link_names[1]);
+        double d = std::max<double>(((dist - dist_result.distance) * coeff), 0.);
+        if (result.gradients[0].has_gradient)
+        {
+          double wd = (d * result.gradients[0].scale);
+          if (wd > err[0])
+            err[0] = wd;
```
mpowelson, line 144: This is like a max weighted by the gradient?

trajopt_ifopt/src/discrete_collision_constraint.cpp
```diff
@@ -0,0 +1,264 @@
...
+    {
+      for (tesseract_collision::ContactResult& dist_result : dist_results)
+      {
+        double dist = collision_evaluator_->GetCollisionConfig().collision_margin_data.getPairCollisionMargin(
+            dist_result.link_names[0], dist_result.link_names[1]);
+        double coeff = collision_evaluator_->GetCollisionConfig().collision_coeff_data.getPairCollisionCoeff(
+            dist_result.link_names[0], dist_result.link_names[1]);
+        err[0] += std::max<double>(((dist - dist_result.distance) * coeff), 0.);
+      }
+      err[0] = err[0] / static_cast<double>(dist_results.size());
```
mpowelson, line 166: This seems to be doing what I would expect for the averaging

trajopt_ifopt/src/discrete_collision_constraint.cpp
```diff
@@ -0,0 +1,264 @@
...
+  switch (gradient_method_)
+  {
+    case GradientCombineMethod::SUM:
+    {
+      grad_vec = getSumGradient(grad_results, n_dof_);
+      break;
+    }
+    case GradientCombineMethod::WEIGHTED_SUM:
+    {
+      grad_vec = getSumGradient(grad_results, n_dof_);
```
mpowelson, line 225: For the weighted, since the value is `e^2` won't the gradient be `2*e*(de/dx)`?

trajopt_optimizers/trajopt_sqp/src/qp_problem.cpp
```diff
@@ -265,42 +302,82 @@ void QPProblem::updateSlackVariableBounds()
...
+    if (constraint_types_[static_cast<std::size_t>(i)] == ConstraintType::EQ)
+    {
+      // The lower and upper should be the same because it is an Equality constraint
+      violation[i] = std::abs(cnt_vals[i] - cnt_bounds[static_cast<std::size_t>(i)].lower_);
+    }
+    else  // ConstraintType::INEQ
+    {
+      // For inequality constraint the valid range is between [0, -inf] so anything greater than
+      // zero is a violation
+      violation[i] = std::max<double>(cnt_vals[i] - cnt_bounds[static_cast<std::size_t>(i)].upper_, 0);
```
mpowelson, line 334: Does this work if the bounds are not 0 to inf? Before, it was checking to see if it was outside either bound

trajopt_ifopt/src/discrete_collision_evaluators.cpp
```diff
@@ -0,0 +1,149 @@
+/**
+ * @file discrete_collision_evaluators.h
```
mpowelson, line 2: Not reviewing this since I assume it was just copied. Let me know if that isn't true

trajopt_ifopt/src/continuous_collision_evaluators.cpp
```diff
@@ -0,0 +1,360 @@
+/**
+ * @file continuous_collision_evaluators.h
```
mpowelson, line 2: Not reviewing this since I assume it was just copied. Let me know if that isn't true

mpowelson (commented):
Ok. Gradually making my way through it. Mostly done except collision_utils, continuous collision, and unit tests.

trajopt_sco/src/osqp_interface.cpp
Levi-Armstrong: Yep, Done.

trajopt_sco/src/optimizers.cpp
Levi-Armstrong: done

trajopt_optimizers/trajopt_sqp/test/test_suite_utils.hpp
Levi-Armstrong: Done.

trajopt_ifopt/include/trajopt_ifopt/constraints/collision_types.h
```diff
@@ -0,0 +1,194 @@
...
+namespace trajopt
+{
+using GetStateFn =
+    std::function<tesseract_environment::EnvState::Ptr(const std::vector<std::string>& joint_names,
+                                                       const Eigen::Ref<const Eigen::VectorXd>& joint_values)>;
+
+/**
+ * @brief IFOPT does not support dynamically changing the constraint size so
+ * all collision gradient data must be combined into a single gradient and error.
+ * This allows switching between different methods for calculating a single gradient and error.
```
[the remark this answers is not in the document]
Levi-Armstrong: Done.

trajopt_ifopt/src/squared_cost.cpp
```diff
@@ -31,23 +31,25 @@ TRAJOPT_IGNORE_WARNINGS_POP
...
+  , n_constraints_(constraint_->GetRows())
   , weights_(weights)
 {
   // Calculate targets - Average the upper and lower bounds
   targets_.resize(n_constraints_);
   std::vector<ifopt::Bounds> bounds = constraint_->GetBounds();
   for (Eigen::Index ind = 0; ind < n_constraints_; ind++)
   {
+    /** @todo Is this correct? (LEVI) */
     targets_(ind) = (bounds[static_cast<std::size_t>(ind)].upper_ + bounds[static_cast<std::size_t>(ind)].lower_) / 2.;
```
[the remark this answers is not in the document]
Levi-Armstrong: It looks good to me now that I am more familiar with the code and how the bounds work.

trajopt_optimizers/trajopt_sqp/include/trajopt_sqp/qp_problem.h
```diff
@@ -79,6 +81,14 @@ class QPProblem
    */
   Eigen::VectorXd evaluateConvexCosts(const Eigen::Ref<const Eigen::VectorXd>& var_vals);
 
+  Eigen::VectorXd evaluateConvexConstraintViolation(const Eigen::Ref<const Eigen::VectorXd>& var_vals);
```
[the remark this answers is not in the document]
Levi-Armstrong: done

trajopt_optimizers/trajopt_sqp/src/qp_problem.cpp
Levi-Armstrong: I believe this was one restriction in the original trajopt that constraint had to be written such that they obeyed this but not sure if we want to impose the same restriction. Well, not that I think about it how did we handle tolerances in the old trajopt?

Levi-Armstrong: I plan to revert to what you had originally. This was mostly to help me track down other issues but what you had originally is what we want.

trajopt_optimizers/trajopt_sqp/src/trust_region_sqp_solver.cpp
```diff
@@ -34,7 +34,7 @@
 
 namespace trajopt_sqp
 {
-const bool SUPER_DEBUG_MODE = false;
+const bool SUPER_DEBUG_MODE = true;
```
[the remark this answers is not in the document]
Levi-Armstrong: done.

trajopt_optimizers/trajopt_sqp/include/trajopt_sqp/trust_region_sqp_solver.h
```diff
@@ -55,10 +55,36 @@ class TrustRegionSQPSolver : public ifopt::Solver
...
+  bool stepSQPSolver();
+
+  /**
+   * @brief Check if the SQPSolver constraints are satisfied
+   * @warning This should not normally be call directly, but exposed for online planning
+   */
+  bool verifySQPSolverConvergence();
+
+  /**
+   * @brief The SQPSolver reported convergence but constarts are not satisfied so this function
```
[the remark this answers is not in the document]
Levi-Armstrong: fixed

trajopt_optimizers/trajopt_sqp/src/trust_region_sqp_solver.cpp
```diff
@@ -179,92 +97,211 @@ void TrustRegionSQPSolver::Solve(ifopt::Problem& nlp)
...
+  else
+  {
+    CONSOLE_BRIDGE_logInform("Not all constraints are satisfied. Increasing constraint penalties uniformly");
+    results_.merit_error_coeffs *= params.merit_coeff_increase_ratio;
+  }
+  results_.box_size = Eigen::VectorXd::Ones(results_.box_size.size()) *
+                      fmax(results_.box_size[0], params.min_trust_box_size / params.trust_shrink_ratio * 1.5);
+}
+
+bool TrustRegionSQPSolver::stepSQPSolver()
```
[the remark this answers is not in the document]
Levi-Armstrong: I think so. Breaking these out into separate functions should make it easier to do things like this for the online planner.

trajopt_ifopt/include/trajopt_ifopt/variable_sets/joint_position_variable.h
```diff
@@ -51,12 +52,35 @@ class JointPosition : public ifopt::VariableSet
...
     values_ = init_value;
   }
 
+  JointPosition(const Eigen::Ref<const Eigen::VectorXd>& init_value,
+                std::vector<std::string> joint_names,
+                const tesseract_common::KinematicLimits& bounds,
+                const std::string& name = "Joint_Position")
+    : ifopt::VariableSet(static_cast<int>(init_value.size()), name), joint_names_(std::move(joint_names))
+  {
+    /** @todo Print warning if init value is not within bounds */
```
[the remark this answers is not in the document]
Levi-Armstrong: Yea that would be good and print a warning.

trajopt_ifopt/src/cartesian_position_constraint.cpp
```diff
@@ -81,11 +81,34 @@ void CartPosConstraint::SetBounds(const std::vector<ifopt::Bounds>& bounds)
...
+
+  // Reserve enough room in the sparse matrix
+  jac_block.reserve(n_dof_ * 6);
+
+  // Determine which rotational error function to use
+  auto fn = &trajopt::calcRotationalError;
+  {
+    Eigen::Isometry3d pose_err = pose_error_calculator(joint_vals);
+    double angle = fn(pose_err.rotation()).norm();
+    // If the able is near M_PI then you can get incorrect jacobians. I near M_PI switch to error functin that returns
```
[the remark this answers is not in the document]
Levi-Armstrong: fixed

trajopt_ifopt/src/discrete_collision_evaluators.cpp
Levi-Armstrong: It was, the only thing I changed is the collision buffer was not being applied to the contact manager and I have added a note asking if we should also add this to the link to link margin (I believe we should but wanted a second opinion).

trajopt_ifopt/src/continuous_collision_evaluators.cpp
Levi-Armstrong: The only thing that has changed other than what is mentioned above, is the hashing of the data.

trajopt_optimizers/trajopt_sqp/src/qp_problem.cpp
Levi-Armstrong: This may also affect the slack variables.

trajopt_optimizers/trajopt_sqp/src/qp_problem.cpp
```diff
@@ -181,39 +187,70 @@ void QPProblem::linearizeConstraints()
...
+  Eigen::VectorXd x_initial = nlp_->GetVariableValues().head(num_nlp_vars_);
+  /** @todo Why not use nlp_->GetConstraints().GetValues()? */
+  Eigen::VectorXd cnt_initial_value = nlp_->EvaluateConstraints(x_initial.data());
+
+  // d = collision distance
+  // T = distance threshhold
+  // J = Jacobian
+  // x = initial values
+  // Original trajopt Affine constant = T - d - (J * x)
+  // Trajopt ifopt cnt_initial_value = T - d, so the following is correct.
```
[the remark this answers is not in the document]
Levi-Armstrong: Removed.

trajopt_optimizers/trajopt_sqp/src/qp_problem.cpp
```diff
@@ -181,39 +187,70 @@ void QPProblem::linearizeConstraints()
...
-    Eigen::VectorXd linearized_cnt_lower = cnt_bound_lower - (cnt_initial_value - jac * x_initial);
-    Eigen::VectorXd linearized_cnt_upper = cnt_bound_upper - (cnt_initial_value - jac * x_initial);
+void QPProblem::updateConstraintsConstantExpression()
+{
+  if (num_nlp_cnts_ == 0)
+    return;
+
+  // Get values about which we will linearize
+  Eigen::VectorXd x_initial = nlp_->GetVariableValues().head(num_nlp_vars_);
+  /** @todo Why not use nlp_->GetConstraints().GetValues()? */
```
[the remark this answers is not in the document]
Levi-Armstrong: Switched.

trajopt_optimizers/trajopt_sqp/src/qp_problem.cpp
```diff
@@ -181,39 +187,70 @@ void QPProblem::linearizeConstraints()
...
+  Eigen::VectorXd x_initial = nlp_->GetVariableValues().head(num_nlp_vars_);
+  Eigen::VectorXd cost_initial_value = nlp_->GetCosts().GetValues();
 
-    // Get values about which we will linearize
-    Eigen::VectorXd x_initial = nlp_->GetVariableValues();
-    Eigen::VectorXd cnt_initial_value = nlp_->EvaluateConstraints(x_initial.data());
+  // The block excludes the slack variables
+  Eigen::VectorXd result_quad = x_initial.transpose() * hessian_.block(0, 0, num_nlp_vars_, num_nlp_vars_) * x_initial;
+  Eigen::VectorXd result_lin = x_initial.transpose() * gradient_.block(0, 0, num_nlp_vars_, num_nlp_costs_);
+  cost_constant_ = cost_initial_value - result_quad - result_lin;
```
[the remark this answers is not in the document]
Levi-Armstrong: Done.

trajopt/include/trajopt/utils.hpp
```diff
@@ -156,25 +156,65 @@ inline std::vector<SafetyMarginData::Ptr> createSafetyMarginDataVector(int num_e
...
-  return r12.axis() * angle;
+  return axis * angle;
+}
+
+/**
+ * @brief Calculate the rotation error vector given a rotation error matrix where the angle is between [0, 2 * pi]
+ * @param R rotation error matrix
+ * @return Rotation error vector = Eigen::AngleAxisd.axis() * Eigen::AngleAxisd.angle()
+ */
+inline Eigen::Vector3d calcRotationalError2(const Eigen::Ref<const Eigen::Matrix3d>& R)
```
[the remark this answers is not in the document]
Levi-Armstrong: Done.

trajopt_ifopt/include/trajopt_ifopt/variable_sets/joint_position_variable.h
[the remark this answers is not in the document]
Levi-Armstrong: Done.

trajopt_optimizers/trajopt_sqp/src/qp_problem.cpp
```diff
@@ -265,42 +302,82 @@ void QPProblem::updateSlackVariableBounds()
...
+  //  Eigen::VectorXd dist_from_lower = -(cnt_bound_lower - cnt_eval);
+  //  Eigen::VectorXd dist_from_upper = cnt_bound_upper - cnt_eval;
+
+  //  // Now we put those values into a matrix
+  //  Eigen::MatrixXd tmp(num_nlp_cnts_, 3);
+  //  tmp << dist_from_lower, dist_from_upper, zero;
+
+  //  // We return the worst violation and flip it so violations are positive
+  //  Eigen::VectorXd violation = -1 * tmp.rowwise().minCoeff();
+  //  return violation;
```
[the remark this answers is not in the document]
Levi-Armstrong: Moved to utility function in trajopt_ifopt `calcBoundsViolations`

trajopt_optimizers/trajopt_sqp/src/qp_problem.cpp
Levi-Armstrong: This has been resolved with the addition of the `calcBoundsErrors` and `calcBoundsViolations`.

Levi-Armstrong:
@mpowelson Added the absolute cost along with unit tests for the current cost wrappers. I think the only remaining thing on this PR is to figure out how best to handle the error calculations for the different methods used to combine the gradient.

trajopt_ifopt/src/absolute_cost.cpp
```diff
@@ -0,0 +1,80 @@
...
+  , constraint_(std::move(constraint))
+  , n_constraints_(constraint_->GetRows())
+  , weights_(weights)
+{
+}
+
+double AbsoluteCost::GetCost() const
+{
+  // This takes the absolute value of the errors
+  Eigen::VectorXd error = calcBoundsViolations(constraint_->GetValues(), constraint_->GetBounds());
```
mpowelson, line 51: Based on the comment in the utils. calcBoundsViolation returns negative values if they violate. So we'll need to flip the sign here.

trajopt_ifopt/src/absolute_cost.cpp
```diff
@@ -0,0 +1,80 @@
...
+  assert(var_size > 0);
+  cnt_jac_block.resize(constraint_->GetRows(), var_size);
+
+  // Get the Jacobian Block from the constraint
+  constraint_->FillJacobianBlock(var_set, cnt_jac_block);
+
+  // Apply the chain rule. See doxygen for this class
+  Eigen::ArrayXd error = calcBoundsErrors(constraint_->GetValues(), constraint_->GetBounds());
+  Eigen::ArrayXd w_error = error * weights_.array();
+  Eigen::VectorXd coeff = w_error / error.abs();
```
mpowelson, line 75: Shouldn't this be `w_error / w_error.abs()`?

mpowelson: Just realized what you are doing. Maybe add a comment here since you are letting the two w's cancel. It is worth noting that this breaks down if w < 0, but that should always be the case if the user sets it up correctly.

trajopt_ifopt/src/cartesian_position_constraint.cpp
```diff
@@ -81,11 +81,23 @@ void CartPosConstraint::SetBounds(const std::vector<ifopt::Bounds>& bounds)
...
+  jac_block.reserve(n_dof_ * 6);
+
   if (use_numeric_differentiation)
   {
-    auto error_calculator = [&](const Eigen::Ref<const Eigen::VectorXd>& x) { return this->CalcValues(x); };
-    Jacobian jac0(6, n_dof_);
-    jac0 = calcForwardNumJac(error_calculator, joint_vals, 1e-5);
+    auto error_calculator = [&](const Eigen::Ref<const Eigen::VectorXd>& x) {
+      Eigen::Isometry3d pose_err = pose_error_calculator(x);
+      Eigen::VectorXd err = concat(pose_err.translation(), calcRotationalError2(pose_err.rotation()));
```
mpowelson, line 97: Why did you remove the other RotationalError calculator and the logic for switching between them?

trajopt_ifopt/src/continuous_collision_constraint.cpp
```diff
@@ -0,0 +1,423 @@
...
+    }
+    case GradientCombineMethod::WEIGHTED_LEAST_SQUARES:
+    {
+      for (tesseract_collision::ContactResult& dist_result : dist_results)
+      {
+        double dist = collision_evaluator_->GetCollisionConfig().collision_margin_data.getPairCollisionMargin(
+            dist_result.link_names[0], dist_result.link_names[1]);
+        double coeff = collision_evaluator_->GetCollisionConfig().collision_coeff_data.getPairCollisionCoeff(
+            dist_result.link_names[0], dist_result.link_names[1]);
+        err[0] += std::max<double>((std::pow(dist - dist_result.distance, 2) * coeff), 0.);
```
mpowelson, line 201: Same here

mpowelson (commented):
Some more comments.

mpowelson:
> @mpowelson Added the absolute cost along with unit tests for the current cost wrappers. I think the only remaining thing on this PR is to figure out how best to handle the error calculations for the different methods used to combine the gradient.

We are trying to reduce a multiple row constraint down to a single row. Thinking about how to reduce these constraints, it seems we currently have 3 options
1) Sum - Makes sense, but it could blow up if you have lots of links (e.g. a dense mesh)
2) Average - Doesn't make sense unless this is absolute value average. You don't want them cancelling out. However, this divides by the number of collisions, so it doesn't blow up when lots of links are in collision at once
3) Max - This also makes sense. I just think it is not the most efficient. It ignores perpendicular axes. If row 1 is only effected by j1 and row 2 is only effected by j2, we should adjust both in the same step. We shouldn't just adjust the row that is bigger.

So I think intuition is that the average violation (which is >0) is the best default value.

Now, we also square some of them currently. This does make sense in that it will weight bigger violations higher. I think we might consider handling this in its own flag or as a couple of functions (or class?) you pass in. One could imagine having other scaling factors as well such as exponential scaling as well.

trajopt_optimizers/trajopt_sqp/src/trust_region_sqp_solver.cpp
```diff
@@ -196,7 +202,7 @@ void TrustRegionSQPSolver::runTrustRegionLoop()
       qp_problem->print();
 
     results_.overall_iteration++;
-    results_.trust_region_iteration = trust_region_iteration;
+    results_.trust_region_iteration = trust_region_iteration + 1;
```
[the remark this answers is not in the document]
Levi-Armstrong: You are correct it was purely for the printout. In the case of penalty iteration, 0 meant that it has not run but in most of the others 0 meant the first iteration. This was to clear things up such that 0 indicates it has not run and now 1 indicates it has run one iteration so things are consistent.

trajopt_ifopt/include/trajopt_ifopt/costs/absolute_cost.h
```diff
@@ -0,0 +1,109 @@
...
+ * 1) Constraints have multiple rows where a cost is simply a value (the sum of all of the costs)
+ * 2) Constraints have bounds that they must be within. Costs should simply be as close to 0 as possible
+ *
+ * This class converts the bounds to a target and then sets the value to be the absolute value of the distances
+ * from that target. The result is that the value (GetCost) is a single double and the jacobian (GetJacobian) is a
+ * single row (1 x n_vars)
+ *
+ * Given a constraint: g(x) with some bounds: upper_ and lower_
+ *
+ * The target of the cost is set to the average of the upper and lower bounds:
```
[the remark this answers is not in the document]
Levi-Armstrong: Updated.

trajopt_ifopt/src/absolute_cost.cpp
Levi-Armstrong: This has been corrected. The function returns the absolute value of the error so it will always be positive.

trajopt_ifopt/src/cartesian_position_constraint.cpp
Levi-Armstrong: After writing the unit tests, the calcRotationalError breaks down when calculating the jacobian at PI or -PI, but the calcRotationalError2 did not break down at any of the rotation angles between -5PI and 5PI when calculating the jacobian. This is because when using Eigen's angle axis it converts the angle to be between [0, PI] where internally if the angle is between [-PI, 0] it flips the sign of the axis. The calcRotationalError both check for this now flip and reverts it and since the angle is always between [-PI, PI] for Eigen Angle axis switching the range to [0, 2PI] will never be close to 2PI. Though I would expect this function to break down near zero it does not appear to but should investigate further.

trajopt_ifopt/src/discrete_collision_constraint.cpp
```diff
@@ -0,0 +1,263 @@
...
+    }
+    case GradientCombineMethod::WEIGHTED_LEAST_SQUARES:
+    {
+      for (tesseract_collision::ContactResult& dist_result : dist_results)
+      {
+        double dist = collision_evaluator_->GetCollisionConfig().collision_margin_data.getPairCollisionMargin(
+            dist_result.link_names[0], dist_result.link_names[1]);
+        double coeff = collision_evaluator_->GetCollisionConfig().collision_coeff_data.getPairCollisionCoeff(
+            dist_result.link_names[0], dist_result.link_names[1]);
+        err[0] += std::max<double>((std::pow(dist - dist_result.distance, 2) * coeff), 0.);
```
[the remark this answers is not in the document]
Levi-Armstrong: Good catch.

trajopt_ifopt/src/continuous_collision_constraint.cpp
Levi-Armstrong: Fixed

trajopt_ifopt/src/absolute_cost.cpp
Levi-Armstrong: Added documentation and updated the constructor to take the absolute value of the weights.

trajopt_ifopt/src/cartesian_position_constraint.cpp
Levi-Armstrong: The reason calcRotationalError2 does not break down at 0 is that we are making sure the angle axis aligns with the quaternion axis which prevents this function from breaking down at zero. 

Angle: 0.001 results in an axis: [0, 0, 1]
Angle: -0.001 results in and axis: [0, 0, -1]
e1 = angle * axis = [0, 0, 0.001]
e2 = angle * axis = [0, 0, -0.001]
delta = e2 - e1 = [0, 0, 0.002]

Levi-Armstrong: Add this information to the documentation.

Levi-Armstrong:
@mpowelson I believe this is ready for review. There are a lot of changes to make the collision evaluators work within the new IFOPT framework. The only remaining change is the current constraints are not excluding the results at the previous and post state which I am working on fixing now but should not change much.

trajopt_ifopt/include/trajopt_ifopt/cache.h
```diff
@@ -0,0 +1,45 @@
+#ifndef TRAJOPT_IFOPT_CACHE_H
```
mpowelson, line 1: Missing license

trajopt_ifopt/include/trajopt_ifopt/costs/absolute_cost.h
```diff
@@ -0,0 +1,109 @@
...
+   * @brief Constructs a CostTerm that converts a constraint into a cost with a sum squared error
+   * @param constraint Input constraint to be converted to a cost
+   */
+  AbsoluteCost(ifopt::ConstraintSet::Ptr constraint);
+  /**
+   * @brief Constructs a CostTerm that converts a constraint into a cost with a weighted sum squared error
+   * @param constraint Input constraint to be converted to a cost
+   * @param weights Weights applied to the constraints. Length should be n_constraints
+   */
+  AbsoluteCost(ifopt::ConstraintSet::Ptr constraint, const Eigen::Ref<const Eigen::VectorXd>& weights);
```
mpowelson, line 90: We had discussed also passing in the nlp so we could register the variables. Did you decide against that?

trajopt_optimizers/trajopt_sqp/src/osqp_eigen_solver.cpp
```diff
@@ -38,7 +38,7 @@ OSQPEigenSolver::OSQPEigenSolver()
   solver_.settings()->setWarmStart(true);
   solver_.settings()->setPolish(true);
   solver_.settings()->setAdaptiveRho(false);
-  solver_.settings()->setMaxIteraction(8192);
+  solver_.settings()->setMaxIteration(8192);
```
mpowelson, line 41: Was this a typo in their api or was this a valid setting that was doing something else?

trajopt_ifopt/src/absolute_cost.cpp
```diff
@@ -0,0 +1,83 @@
...
+AbsoluteCost::AbsoluteCost(ifopt::ConstraintSet::Ptr constraint)
+  : AbsoluteCost(std::move(constraint), Eigen::VectorXd::Ones(constraint->GetRows()))
+{
+}
+
+AbsoluteCost::AbsoluteCost(ifopt::ConstraintSet::Ptr constraint, const Eigen::Ref<const Eigen::VectorXd>& weights)
+  : CostTerm(constraint->GetName() + "_absolute_cost")
+  , constraint_(std::move(constraint))
+  , n_constraints_(constraint_->GetRows())
+  , weights_(weights.cwiseAbs())  // must be positive
```
mpowelson, line 44: Ideally I think we should throw an error if weights is the wrong size. Otherwise I think Eigen will throw in FillJacobianBlock with a probably more cryptic error

trajopt_ifopt/include/trajopt_ifopt/utils/ifopt_utils.h
```diff
@@ -111,5 +112,111 @@ inline Eigen::VectorXd getClosestValidPoint(const Eigen::Ref<const Eigen::Vector
...
+ * @param bounds The bounds
+ * @return The absolute errors given the bounds, if within the bounds it will be zero
+ */
+inline Eigen::VectorXd calcBoundsViolations(const Eigen::Ref<const Eigen::VectorXd>& input,
+                                            const std::vector<ifopt::Bounds>& bounds)
+{
+  return calcBoundsErrors(input, bounds).cwiseAbs();
+}
+
+inline ifopt::Problem::VectorXd calcNumericalCostGradient(const double* x, ifopt::Problem& nlp, double epsilon = 1e-8)
```
mpowelson, line 156: Could consider pushing this back to ifopt in the future.

mpowelson (commented):
Another partial review. Still need to look at collisino

mpowelson (commented):
You might consider putting collision in its own directory. I still need to look at the collision utils again. Are those in the final format you were going to put them in?

trajopt_optimizers/trajopt_sqp/src/osqp_eigen_solver.cpp
Levi-Armstrong: I think so on their end so it was deprecated.

Levi-Armstrong:
@mpowelson The continuous IFOPT constraint is now working and should be ready for you to review.

PR #281, opened 2021-12-08
On final convex solver failure attempt set trust region to minimum

Levi-Armstrong:


mpowelson (approved):
Looks good. It is worth noting that this is a slight change in behavior. Now solver failures don't reset if it converges once. You get 3 solver failures for the entire optimization. 

This also no longer re-convexifies in between solver failures. I only adjusts the box constraints. Which I think is what we want.

PR #282, opened 2021-12-10
Use applyContactManagerConfig instead of manually setting

mpowelson:
Switch to using applyContactManagerConfig rather than doing it manually (and partially)

### DIFF
--- a/trajopt_ifopt/src/constraints/collision/continuous_collision_evaluators.cpp
+++ b/trajopt_ifopt/src/constraints/collision/continuous_collision_evaluators.cpp
@@ -68,11 +68,7 @@ LVSContinuousCollisionEvaluator::LVSContinuousCollisionEvaluator(
 
   contact_manager_ = env_->getContinuousContactManager();
   contact_manager_->setActiveCollisionObjects(manip_active_link_names_);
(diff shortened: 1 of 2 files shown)


mpowelson:
This is failing because it doesn't handle the collision buffer. I'll have to look more into what to do here. There is also interplay with the CollisionCoeffData

Levi-Armstrong:
I think to apply the buffer it should probably internally create a new margin data by looping over the provided one and increasing the margin buffer. Though we would need to make sure that pairs with custom distance other than the default are handled correctly with regards to the buffer.

Levi-Armstrong:
@mpowelson what are the next steps on this PR?

mpowelson:
> @mpowelson what are the next steps on this PR?

I'm not sure. We'll have to figure out how we want to handle the CollisionCoeffData and the Collision margin data. Maybe we should add an apply method to the config. Otherwise, I'm not sure how to make the contact manager apply derived configs like this correctly.

drchrislewis:
@Levi-Armstrong https://m.youtube.com/watch?v=v6MhPl2ICsc don't have your current email. Saw this video, seems like control of walking is a real time nonlinear constrained optimization problem using distance field.

PR #172, opened 2020-03-03
Collision Interpolation Bug Fix

marip8:
This PR fixes a bug where the `cc_time` of the second link in a collision pair was being indexed incorrectly. It also replaces code with for loop since the same operations are applied to both collision links

PR #285, opened 2022-03-24
Expose convex solver settings and set ospq adaptive_rho to default value

Levi-Armstrong:
The adaptive_rho was change because a unit test was failing [here](https://github.com/tesseract-robotics/trajopt/pull/172) but now passes using the default setting indicating that either there was an issue in OSQP or TrajOpt that has been fixed. Have adaptive_rho enable has a signification impact on planning time so it is reverted to default value now that the setting is exposed.

### DIFF
--- a/.github/workflows/windows_noetic_build.yml
+++ b/.github/workflows/windows_noetic_build.yml
@@ -12,7 +12,7 @@ on:
 jobs:
   windows_ci:
     name: Noetic
-    runs-on: windows-latest
+    runs-on: windows-2019
     env:
       ROS_DISTRO: noetic
     steps:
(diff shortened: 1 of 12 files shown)


Levi-Armstrong:
@marrts I made some slight modification to your changes. Thanks for the improvement.

mpowelson:
Relevant PR when adaptive rho was disabled: https://github.com/tesseract-robotics/trajopt/pull/172

mpowelson (approved):
Look reasonable. Someone has run some tests and determined that this is a better default?

marrts:
> Look reasonable. Someone has run some tests and determined that this is a better default?

Yes, this went from failing from iterations, which I found needed close to 10 million iterations to solving in about 10k iterations. Substantial speed improvement

PR #288, opened 2022-04-22
Protect jsoncpp against multiple find_package() calls

johnwason:
There is a bug in newer `jsoncpp` versions that raises an error if `find_package(jsoncpp)` is called more than once. This PR fences the calls to check if the offending target already exists.

See https://github.com/open-source-parsers/jsoncpp/issues/1356


trajopt/CMakeLists.txt
```diff
@@ -13,8 +13,11 @@ find_package(trajopt_utils REQUIRED)
 find_package(tesseract_common REQUIRED)
 find_package(tesseract_environment REQUIRED)
 find_package(tesseract_visualization REQUIRED)
-find_package(jsoncpp REQUIRED)
+if(NOT TARGET JsonCpp::JsonCpp)
+  find_package(jsoncpp REQUIRED)
+endif()
 find_package(ros_industrial_cmake_boilerplate REQUIRED)
+find_package(osqp REQUIRED)
```
Levi-Armstrong, line 20: Should this be added.

johnwason: Sorry that was from something else and I didn't remove it.

Levi-Armstrong (approved):

PR #290, opened 2022-06-28
CPack

marip8:
This PR adds CPack to the packages in this repository. I've tested these changes with Ubuntu Focal, but I haven't yet tested with Windows. I also used the CMake boilerplate tools to extract package names and versions


Levi-Armstrong (changes requested):
Over all looks good. Thanks for knocking this out.

PR #291, opened 2022-06-29
Support storing sco::Optimizer::Callbacks in ProblemConstructionInfo class

Levi-Armstrong:


Levi-Armstrong:
@johnwason Any thoughts on why the windows build is finding a cache but seems to still build all of the vcpkgs? This does not happen on the tesseract build.

johnwason:
@Levi-Armstrong no, I don't see where the issues is. I don't know if it will cache a pull request, maybe that is the issue?

Levi-Armstrong:
> @Levi-Armstrong no, I don't see where the issues is. I don't know if it will cache a pull request, maybe that is the issue?

Is that what this line is doing?
`if: startsWith(github.ref, 'refs/tags/v') != true`

johnwason:
No, that line prevents caching for release builds. I start my releases with `v`.

PR #289, opened 2022-04-22
coef.sparseView() is failing on newer versions of eigen in absolute_c…

johnwason:
`coef.sparseView()` is failing to compile on newer versions of Eigen. This PR removes the sparseView().

Issue #298, opened 2022-09-05
jsoncpp includes cause linking errors in tesseract_planning


rjoomen:
The way jsoncpp is included in the CMake target_link_libraries causes a linking error in tesseract_planning on Ubuntu 22.04:
`/usr/bin/ld: cannot find -ljsoncpp_lib: No such file or directory`

rjoomen:
For reference: open-source-parsers/jsoncpp#1356

Levi-Armstrong:
Thanks for the fix.

PR #299, opened 2022-09-05
Fix jsoncpp link targets and sparseview issue

rjoomen:
Fixes #289 and #298


rjoomen:
Accidentally added two fixes to one PR, sorry.

Levi-Armstrong:
It looks like the target is not present on Focal. I recommend doing the same things we are doing for console_bridge above in the cmake where we check for the target and if it does not exist we create one. This will also need to be added to the cmake config in the cmake folder.

Levi-Armstrong:
@rjoomen I think this should be the following. Is there a reason why we are not always calling find_dependency; because I believe this is why we are having the global scope issue.
```
 find_dependency(jsoncpp)
 if(NOT TARGET jsoncpp_lib)
  add_library(jsoncpp_lib ALIAS JsonCpp::JsonCpp)
```

PR #302, opened 2023-02-03
Add parallelization

marip8:
Various changes from @smith-doug to speed up Trajopt

### DIFF
--- a/trajopt/include/trajopt/collision_terms.hpp
+++ b/trajopt/include/trajopt/collision_terms.hpp
@@ -1,4 +1,5 @@
 #pragma once
+#include <mutex>
 #include <tesseract_environment/environment.h>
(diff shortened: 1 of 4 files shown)


Levi-Armstrong:
The only thing I would request is that the number of threads used be configurable similar to descartes.

marip8:
Any reason why CI fails on the Trajopt JSON stuff? Seems way outside the additions of this PR

marip8:
> The only thing I would request is that the number of threads used be configurable similar to descartes.

How do you suggest doing this for the `trajopt_sco` additions that were made to static functions?

Levi-Armstrong:
It was most likely a result of the mutex header being added.

Levi-Armstrong:
I have run into this before when adding headers, and had to go through and fix the issues.

PR #304, opened 2023-02-21
catkin dependency for ROS1 only

### LINKED ISSUE #303: catkin dependencies
`trajopt_sco, trajopt_ifopt, trajopt_sqp, trajopt_utils` currently have a dependency on `catkin`. 

Can this be removed, as 1) trajopt is ROS-agnostic, and 2) if used with ROS2, catkin is not correct anyway.

rjoomen:
Fixes #303

### DIFF
--- a/trajopt_ifopt/package.xml
+++ b/trajopt_ifopt/package.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0"?>
-<package format="2">
+<package format="3">
   <name>trajopt_ifopt</name>
   <version>0.3.1</version>
   <description>The trajopt_ifopt package</description>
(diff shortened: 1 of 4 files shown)


marip8:
I think all the `<exec_depend>catkin</exec_depend>` can be completely removed. All of these packages are independent of ROS and catkin, and I don't see any macros or commands in the CMakeLists files that come from `catkin`. I think these are just remnants of the original development that was done in ROS1

gavanderhoorn:
REP 136 actually recommends adding that `exec_depend`. I don't remember exactly why, but I believe it has something to do with environment setup files otherwise not getting generated, which would make include, linker and cfg paths not get setup.

marip8:
I see. Disregard my previous comment then

gavanderhoorn:
Just repeating what the REP suggests. If we don't need that the `exec_depend` could be removed.

Levi-Armstrong (approved):

Levi-Armstrong:
> Just repeating what the REP suggests. If we don't need that the `exec_depend` could be removed.

I tried to find more information on exactly why this was needed, but could not find much. I removed these from Tesseract repository a while back and does not appear to have any issues. There is a current issue where if you have tesseract in your workspace and installed it will not use the ones in your package for some of the packages. I thought this was due to some things we had to change to make the windows build work but those have been removed. Maybe this has something to do with it.

PR #323, opened 2023-05-24
MSVC 2022 build fixes

johnwason:
This PR adds MSVC 2022 CI build, and fixes Eigen Index to long conversion issues.

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -19,7 +19,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [windows-2019]
+        os: [windows-2019,windows-2022]
     steps:
     - uses: actions/checkout@v3
       with:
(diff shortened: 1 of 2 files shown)


dependencies.repos
```diff
@@ -5,7 +5,7 @@
 - git:
     local-name: tesseract
     uri: https://github.com/ros-industrial-consortium/tesseract.git
-    version: 0.16.1
+    version: 2a50d0f458f4f074ed465999521da51dc7159c30
```
Levi-Armstrong, line 8: Lets use the master branch.

johnwason: I changed it to master. We need a new tag with the fixes in that repo at some point.

johnwason:
I am seeing some test failures. I will take a look why when I have the change, but this should be ok to merge with those failures on Windows.

PR #333, opened 2023-06-19
Add multi-threaded support to TrajOpt

Levi-Armstrong:
This replaces #302 and surprised that it was not having issues because the Models themselves needs to be threadsafe. This is why that PR required the omp critical around the addHinge, but does not solve the issue where other constraints are also adding variables and other items to the model which was not threadsafe. 

@marip8 I came up with a way to separate running trajopt with and without multi-threading but not sure if there is a better way.

### DIFF
--- a/trajopt_sco/include/trajopt_sco/optimizers.hpp
+++ b/trajopt_sco/include/trajopt_sco/optimizers.hpp
@@ -119,7 +119,69 @@ struct BasicTrustRegionSQPParameters
   bool log_results;     // Log results to file
   std::string log_dir;  // Directory to store log results (Default: /tmp)
 
+  /** @brief If greater than one, multi threaded functions are called */
+  int num_threads;
+
   BasicTrustRegionSQPParameters();
+};
+
+struct BasicTrustRegionSQPUtilFunctions
+{
+  virtual ~BasicTrustRegionSQPUtilFunctions() = default;
+
+  virtual DblVec evaluateCosts(const std::vector<Cost::Ptr>& costs, const DblVec& x) const;
+
+  virtual DblVec evaluateConstraintViols(const std::vector<Constraint::Ptr>& constraints, const DblVec& x) const;
+
+  virtual std::vector<ConvexObjective::Ptr> convexifyCosts(const std::vector<Cost::Ptr>& costs,
+                                                           const DblVec& x,
+                                                           Model* model) const;
+
+  virtual std::vector<ConvexConstraints::Ptr> convexifyConstraints(const std::vector<Constraint::Ptr>& cnts,
+                                                                   const DblVec& x,
+                                                                   Model* model) const;
+
+  virtual DblVec evaluateModelCosts(const std::vector<ConvexObjective::Ptr>& costs, const DblVec& x) const;
+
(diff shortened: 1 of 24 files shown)


Levi-Armstrong:
@marip8 Another option is make these methods virtual method of BasicTrustRegionSQP and then create BasicTrustRegionSQPOMP which overloads these methods. I think this approach is cleaner what do you think?

Levi-Armstrong:
Alright I decided to leverage inheritance which I think is cleaner. Also added multiple unit tests and benchmark.

Levi-Armstrong:
I ran the benchmarks and heaptrack to confirm the changes to not impact the default behavior. The multi threaded approach has a marginal improvement for the planning unit test and memory is doubled based on heaptrack. This is most likely due to each of the operation are not leveraging the same thread pool so the thread local caching is not as effective. In tesseract_planning I will implement a version which leverage the executor so it may perform better.

trajopt_sco/include/trajopt_sco/optimizers.hpp
```diff
@@ -119,9 +119,154 @@ struct BasicTrustRegionSQPParameters
   bool log_results;     // Log results to file
   std::string log_dir;  // Directory to store log results (Default: /tmp)
 
+  /** @brief If greater than one, multi threaded functions are called */
+  int num_threads;
+
   BasicTrustRegionSQPParameters();
 };
 
+struct BasicTrustRegionSQPUtilFunctions
```
marip8, line 128: What's the purpose of this struct and the threaded version? I don't see that they get used anywhere

marip8:
> The multi threaded approach has a marginal improvement for the planning unit test and memory is doubled based on heaptrack. 

Interesting. How hard are these unit test problems (number of waypoints, constraints, etc.)? I would have thought there would be a significant time improvement

I haven't had a chance to look through this thoroughly yet, but it seems like a reasonable approach

trajopt_sco/include/trajopt_sco/optimizers.hpp
Levi-Armstrong: I switched to the inheritance approach because it allows implementing different threading libraries. For example I will create an implementation that uses the task composer executor so the same thread pool can be used.

Levi-Armstrong:
> > The multi threaded approach has a marginal improvement for the planning unit test and memory is doubled based on heaptrack. 
> 
> Interesting. How hard are these unit test problems (number of waypoints, constraints, etc.)? I would have thought there would be a significant time improvement

The planning unit test takes like 23 seconds but only has 6 constraints and 8 costs. I think this is a small problem, but the other issue is I believe each use of omp has its own thread pool so the caching is not as effective hindering performance.

trajopt_sco/include/trajopt_sco/optimizers.hpp
Levi-Armstrong: Those were left over from a previous attempt. They have been removed.

Levi-Armstrong:
I will test it on the puzzle piece example.

Levi-Armstrong:
After CI finishes I will merge, but it would be good if you all have a test case to try and report back.

Levi-Armstrong:
I tried the puzzle piece example but it also showed no difference. I the threads ramp up but does not make a difference for these use cases.

PR #341, opened 2023-07-30
Fix TrustRegionSQPSolver solve method to correctly set status

Levi-Armstrong:
@rjoomen Do you mind reviewing this?

### DIFF
--- a/trajopt_optimizers/trajopt_sqp/src/trust_region_sqp_solver.cpp
+++ b/trajopt_optimizers/trajopt_sqp/src/trust_region_sqp_solver.cpp
@@ -78,7 +78,7 @@ void TrustRegionSQPSolver::solve(const QPProblem::Ptr& qp_problem)
   status_ = SQPStatus::RUNNING;
 
   // Start time
-  using Clock = std::chrono::high_resolution_clock;
+  using Clock = std::chrono::steady_clock;
   auto start_time = Clock::now();
 
   // Initialize solver
@@ -101,15 +101,15 @@ void TrustRegionSQPSolver::solve(const QPProblem::Ptr& qp_problem)
         break;
       }
 
-      if (stepSQPSolver())
-        break;
-
       if (results_.overall_iteration >= params.max_iterations)
       {
         CONSOLE_BRIDGE_logInform("Iteration limit");
         status_ = SQPStatus::ITERATION_LIMIT;
         break;
       }
+
+      if (stepSQPSolver())
+        break;
     }
 
     // Check if constraints are satisfied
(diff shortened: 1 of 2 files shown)


trajopt_optimizers/trajopt_sqp/src/trust_region_sqp_solver.cpp
```diff
@@ -88,6 +93,14 @@ void TrustRegionSQPSolver::solve(const QPProblem::Ptr& qp_problem)
     // Convexification loop
     for (int convex_iteration = 0; convex_iteration < 100; convex_iteration++)
     {
+      double elapsed_time = std::chrono::duration<double, std::milli>(Clock::now() - start_time).count() / 1000.0;
```
rjoomen, line 96: You can convert to std::seconds here and remove the division by 1000.0

trajopt_optimizers/trajopt_sqp/src/trust_region_sqp_solver.cpp
```diff
@@ -88,6 +93,14 @@ void TrustRegionSQPSolver::solve(const QPProblem::Ptr& qp_problem)
...
+      if (elapsed_time > params.max_time)
+      {
+        CONSOLE_BRIDGE_logInform("Elapsed time %f has exceeded max time %f", elapsed_time, params.max_time);
+        status_ = SQPStatus::TIME_LIMIT;
+        break;
+      }
+
       if (stepSQPSolver())
         break;
```
rjoomen, line 106: Maybe also move the iteration limit check (line 107) to before stepSQPSolver(), otherwise in a hypothetical case where the very last solve step succeeds, but the iteration limit is reached at the same time, the whole solve process fails.

trajopt_optimizers/trajopt_sqp/src/trust_region_sqp_solver.cpp
```diff
@@ -76,6 +77,10 @@ void TrustRegionSQPSolver::solve(const QPProblem::Ptr& qp_problem)
 {
   status_ = SQPStatus::RUNNING;
 
+  // Start time
+  using Clock = std::chrono::high_resolution_clock;
```
rjoomen, line 81: Use std::chrono::steady_clock instead, as usage of [high_resolution_clock](https://en.cppreference.com/w/cpp/chrono/high_resolution_clock) is discouraged (see Notes section of link).

rjoomen (changes requested):
Looks good to me, apart from a few small comments, thanks for the quick solution!

trajopt_optimizers/trajopt_sqp/src/trust_region_sqp_solver.cpp
Levi-Armstrong: fixed

Levi-Armstrong:
The unit is not going to pass for the numerical ik because this exposed potential issue which will be addressed in another PR.

PR #342, opened 2023-07-31
Fix ifopt continuous collision evaluator to distinguish CONTINUOUS and LVS_CONTINUOUS

Levi-Armstrong:
Address this [issue](https://github.com/tesseract-robotics/tesseract_planning/issues/355)

### DIFF
--- a/trajopt_ifopt/src/constraints/collision/continuous_collision_evaluators.cpp
+++ b/trajopt_ifopt/src/constraints/collision/continuous_collision_evaluators.cpp
@@ -39,6 +39,10 @@ LVSContinuousCollisionEvaluator::LVSContinuousCollisionEvaluator(
   , collision_config_(std::move(collision_config))
   , dynamic_environment_(dynamic_environment)
 {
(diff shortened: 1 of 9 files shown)


Levi-Armstrong:
@rjoomen FYI

trajopt_ifopt/src/constraints/collision/continuous_collision_evaluators.cpp
```diff
@@ -39,6 +39,10 @@ LVSContinuousCollisionEvaluator::LVSContinuousCollisionEvaluator(
   , collision_config_(std::move(collision_config))
   , dynamic_environment_(dynamic_environment)
 {
+  if (collision_config_->type != tesseract_collision::CollisionEvaluatorType::CONTINUOUS &&
```
rjoomen, line 42: Is this check necessary? You don't do this in the LVSDiscreteCollisionEvaluator or the SingleTimestepCollisionEvaluator either. Not that it's bad to have an extra check, but for performance reasons. Or add the check everywhere.

PR #343, opened 2023-07-31
Fix osqp_eigen_solver hessian and linear constraint matrix.

Levi-Armstrong:
There were several issues found

1. Not correctly using eigen prune
2. Not using pruned matrix in one case 
3. OSQP Eigen uses a different method for populating the CSC Matrix which has a bug if the input is an empty sparse matrix so a work around was implemented to get around the issue until it is fixed upstream
4. Improve debug output to easily compare between Ifopt and Legacy Version


Levi-Armstrong:
@rjoomen FYI

PR #350, opened 2023-10-02
Move TrajOptIfopt collision gradient types and utils to trajopt_common package

Levi-Armstrong:
@rjoomen


rjoomen:
Looks good! I've done some testing, and this works with the Ifopt version of the basic_cartesian_example. But when fixing a mistake in my [Ifopt adaptation of the car_seat_example](https://github.com/tesseract-robotics/tesseract_planning/pull/389), I got a crash, both in this branch and in the old version. I have tracked it down to [this line](https://github.com/tesseract-robotics/trajopt/blob/master/trajopt_optimizers/trajopt_sqp/src/trajopt_qp_problem.cpp#L727), where the `hinge_costs_.GetValues()` fails for `"LVSDiscreteCollision_10"`. I'm unsure what causes this, though.

rjoomen:
I think you can merge this. And tesseract_planning needs only minor modifications to support this.

PR #354, opened 2023-11-03
Add toleranced Cartesian waypoints to solver

marrts:
This is to support tolerances on Cartesian waypoints so that you can set constraints that don't have to match perfectly

### DIFF
--- a/trajopt/include/trajopt/kinematic_terms.hpp
+++ b/trajopt/include/trajopt/kinematic_terms.hpp
@@ -16,6 +16,8 @@ TRAJOPT_IGNORE_WARNINGS_POP
 
 namespace trajopt
 {
+const double DEFAULT_EPSILON = 1e-5;
+
 using ErrorFunctionType = std::function<Eigen::VectorXd(const Eigen::Isometry3d&, const Eigen::Isometry3d&)>;
 
 /**
@@ -96,6 +98,9 @@ struct DynamicCartPoseJacCalculator : sco::MatrixOfVector
    */
   Eigen::VectorXi indices_;
 
+  /** @brief perturbation amount for calculating Jacobian */
+  double epsilon_;
+
   DynamicCartPoseJacCalculator(
       tesseract_kinematics::JointGroup::ConstPtr manip,
       std::string source_frame,
@@ -109,6 +114,7 @@ struct DynamicCartPoseJacCalculator : sco::MatrixOfVector
     , target_frame_(std::move(target_frame))
     , target_frame_offset_(target_frame_offset)
     , indices_(std::move(indices))
+    , epsilon_(DEFAULT_EPSILON)
   {
     assert(indices_.size() <= 6);
   }
@@ -197,6 +203,9 @@ struct CartPoseJacCalculator : sco::MatrixOfVector
    */
   Eigen::VectorXi indices_;
 
+  /** @brief perturbation amount for calculating Jacobian */
+  double epsilon_;
+
   CartPoseJacCalculator(
       tesseract_kinematics::JointGroup::ConstPtr manip,
       std::string source_frame,
@@ -210,6 +219,7 @@ struct CartPoseJacCalculator : sco::MatrixOfVector
     , target_frame_(std::move(target_frame))
     , target_frame_offset_(target_frame_offset)
     , indices_(std::move(indices))
+    , epsilon_(DEFAULT_EPSILON)
   {
     is_target_active_ = manip_->isActiveLinkName(target_frame_);
     assert(indices_.size() <= 6);
--- a/trajopt/include/trajopt/kinematic_terms.hpp
+++ b/trajopt/include/trajopt/kinematic_terms.hpp
@@ -45,7 +45,7 @@ struct DynamicCartPoseErrCalculator : public TrajOptVectorOfVector
 
   /** @brief Error function for calculating the error in the position given the source and target positions
    * this defaults to tesseract_common::calcTransformError if unset*/
-  ErrorFunctionType error_function = nullptr;
+  ErrorFunctionType error_function{ nullptr };
 
   /**
    * @brief This is a vector of indices to be returned Default: {0, 1, 2, 3, 4, 5}
@@ -62,8 +62,8 @@ struct DynamicCartPoseErrCalculator : public TrajOptVectorOfVector
       const Eigen::Isometry3d& source_frame_offset = Eigen::Isometry3d::Identity(),
       const Eigen::Isometry3d& target_frame_offset = Eigen::Isometry3d::Identity(),
       Eigen::VectorXi indices = Eigen::Matrix<int, 1, 6>(std::vector<int>({ 0, 1, 2, 3, 4, 5 }).data()),
-      Eigen::VectorXd lower_tolerance = Eigen::VectorXd::Zero(6),
-      Eigen::VectorXd upper_tolerance = Eigen::VectorXd::Zero(6));
+      Eigen::VectorXd lower_tolerance = {},
+      Eigen::VectorXd upper_tolerance = {});
 
   void Plot(const tesseract_visualization::Visualization::Ptr& plotter, const Eigen::VectorXd& dof_vals) override;
 
@@ -148,7 +148,7 @@ struct CartPoseErrCalculator : public TrajOptVectorOfVector
 
   /** @brief Error function for calculating the error in the position given the source and target positions
    * this defaults to tesseract_common::calcTransformError if unset*/
-  ErrorFunctionType error_function_ = nullptr;
+  ErrorFunctionType error_function_{ nullptr };
 
   /**
    * @brief This is a vector of indices to be returned Default: {0, 1, 2, 3, 4, 5}
@@ -165,8 +165,8 @@ struct CartPoseErrCalculator : public TrajOptVectorOfVector
       const Eigen::Isometry3d& source_frame_offset = Eigen::Isometry3d::Identity(),
       const Eigen::Isometry3d& target_frame_offset = Eigen::Isometry3d::Identity(),
       Eigen::VectorXi indices = Eigen::Matrix<int, 1, 6>(std::vector<int>({ 0, 1, 2, 3, 4, 5 }).data()),
-      Eigen::VectorXd lower_tolerance = Eigen::VectorXd::Zero(6),
-      Eigen::VectorXd upper_tolerance = Eigen::VectorXd::Zero(6));
+      Eigen::VectorXd lower_tolerance = {},
+      Eigen::VectorXd upper_tolerance = {});
 
   void Plot(const tesseract_visualization::Visualization::Ptr& plotter, const Eigen::VectorXd& dof_vals) override;
 
--- a/trajopt/include/trajopt/kinematic_terms.hpp
+++ b/trajopt/include/trajopt/kinematic_terms.hpp
@@ -148,7 +148,7 @@ struct CartPoseErrCalculator : public TrajOptVectorOfVector
 
   /** @brief Error function for calculating the error in the position given the source and target positions
    * this defaults to tesseract_common::calcTransformError if unset*/
-  ErrorFunctionType error_function = nullptr;
+  ErrorFunctionType error_function_ = nullptr;
 
   /**
    * @brief This is a vector of indices to be returned Default: {0, 1, 2, 3, 4, 5}
--- a/trajopt/include/trajopt/kinematic_terms.hpp
+++ b/trajopt/include/trajopt/kinematic_terms.hpp
@@ -194,13 +194,13 @@ struct CartPoseErrCalculator : public TrajOptVectorOfVector
       Eigen::VectorXd lower_tolerance = Eigen::VectorXd::Zero(6),
       Eigen::VectorXd upper_tolerance = Eigen::VectorXd::Zero(6),
       ErrorFunctionType error_func = nullptr)
-      : manip_(std::move(manip))
-      , source_frame_(std::move(source_frame))
-      , target_frame_(std::move(target_frame))
-      , source_frame_offset_(source_frame_offset)
-      , target_frame_offset_(target_frame_offset)
-      , indices_(std::move(indices))
-      , error_function(error_func)
+    : manip_(std::move(manip))
+    , source_frame_(std::move(source_frame))
+    , target_frame_(std::move(target_frame))
+    , source_frame_offset_(source_frame_offset)
+    , target_frame_offset_(target_frame_offset)
+    , indices_(std::move(indices))
+    , error_function(error_func)
   {
     if (lower_tolerance.size() == 0)
       lower_tolerance_ = Eigen::VectorXd::Zero(6);
--- a/trajopt/include/trajopt/kinematic_terms.hpp
+++ b/trajopt/include/trajopt/kinematic_terms.hpp
@@ -3,6 +3,7 @@
 TRAJOPT_IGNORE_WARNINGS_PUSH
 #include <Eigen/Core>
 
+#include <trajopt_common/utils.hpp>
 #include <tesseract_environment/environment.h>
 #include <tesseract_environment/utils.h>
 #include <tesseract_kinematics/core/joint_group.h>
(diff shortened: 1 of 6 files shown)


trajopt/include/trajopt/kinematic_terms.hpp
```diff
@@ -52,14 +59,24 @@ struct DynamicCartPoseErrCalculator : public TrajOptVectorOfVector
       std::string target_frame,
       const Eigen::Isometry3d& source_frame_offset = Eigen::Isometry3d::Identity(),
       const Eigen::Isometry3d& target_frame_offset = Eigen::Isometry3d::Identity(),
-      Eigen::VectorXi indices = Eigen::Matrix<int, 1, 6>(std::vector<int>({ 0, 1, 2, 3, 4, 5 }).data()))
+      Eigen::VectorXi indices = Eigen::Matrix<int, 1, 6>(std::vector<int>({ 0, 1, 2, 3, 4, 5 }).data()),
+      Eigen::VectorXd lower_tolerance = Eigen::VectorXd::Zero(6),
```
marip8, line 63: Default values should probably be something small but non-zero like `Eigen::VectorXd::Constant(6, 1.0e-6)`

trajopt/src/kinematic_terms.cpp
```diff
@@ -29,16 +29,49 @@ using Eigen::VectorXd;
...
+    // If error is above upper tolerance, set resultant to error - upper_tolerance
+    else if (error(i) > upper_tolerance(i))
+    {
+      resultant(i) = error(i) - upper_tolerance(i);
+    }
+  }
+
+  return resultant;
+}
+
```
marip8, line 61: I think this already exists in Trajopt IFOPT in a vectorized, potentially faster way [here](https://github.com/tesseract-robotics/trajopt/blob/2d965b931015fba9a83ca697e114e3c07ef7c25b/trajopt_ifopt/src/utils/ifopt_utils.cpp#L119-L125)

trajopt/include/trajopt/kinematic_terms.hpp
Levi-Armstrong: Why?

Levi-Armstrong:
In order for this to not have stability issues this cannot use the the jacobian utility to provide the jacobain because the error function is not continuous. When within the tolerance this will produce an all zero jacobian which is not correct so this will need to have its own jacobian method. This need to more or less use the error function without tolerance to calculate the jacobian.

marrts:
> In order for this to not have stability issues this cannot use the the jacobian utility to provide the jacobain because the error function is not continuous. When within the tolerance this will produce an all zero jacobian which is not correct so this will need to have its own jacobian method. This need to more or less use the error function without tolerance to calculate the jacobian.

This is why in my tesseract_ planning PR I made it so that we can simultaneously have both costs and constraints, so the cost will continue to drive it to zero while the constraint will keep it inside the bounds. However, if there's a better solution I'd definitely be open to making the changes for that.

Levi-Armstrong:
I would update this so the class holds a std::function which should be used to calculate the error. Then based on construction if toleranced we point it to one function and if not tolerance it points to another function.

marrts:
> I would update this so the class holds a std::function which should be used to calculate the error. Then based on construction if toleranced we point it to one function and if not tolerance it points to another function. The function have the signature of `calcTransformError` -> `VectorXd err = tesseract_common::calcTransformError(target_tf, source_tf);`

Does this have anything to do with the Jacobian issue, or is the purpose of this to save on computation in the cases where there is no tolerance applied?

trajopt/src/problem_description.cpp
```diff
@@ -777,12 +777,24 @@ void DynamicCartPoseTermInfo::hatch(TrajOptProb& prob)
...
+                                                            source_frame_offset,
+                                                            target_frame_offset,
+                                                            indices,
+                                                            lower_tolerance,
+                                                            upper_tolerance);
 
     // This is currently not being used. There is an intermittent bug that needs to be tracked down it is not used.
-    auto dfdx = std::make_shared<DynamicCartPoseJacCalculator>(
-        prob.GetKin(), source_frame, target_frame, source_frame_offset, target_frame_offset, indices);
+    auto dfdx = std::make_shared<DynamicCartPoseJacCalculator>(prob.GetKin(),
```
Levi-Armstrong, line 790: The jacobian should not take into account tolerances.

trajopt/src/problem_description.cpp
```diff
@@ -920,23 +932,47 @@ void CartPoseTermInfo::hatch(TrajOptProb& prob)
...
+                                                     source_frame_offset,
+                                                     target_frame_offset,
+                                                     indices,
+                                                     lower_tolerance,
+                                                     upper_tolerance);
 
     // This is currently not being used. There is an intermittent bug that needs to be tracked down it is not used.
-    auto dfdx = std::make_shared<CartPoseJacCalculator>(
-        prob.GetKin(), source_frame, target_frame, source_frame_offset, target_frame_offset, indices);
+    auto dfdx = std::make_shared<CartPoseJacCalculator>(prob.GetKin(),
```
Levi-Armstrong, line 945: Same here.

trajopt/src/problem_description.cpp
```diff
@@ -920,23 +932,47 @@ void CartPoseTermInfo::hatch(TrajOptProb& prob)
...
```
Levi-Armstrong, line 968: Same here.

Levi-Armstrong (changes requested):
Also need to figure out how to handle the jacobian issue when toleranced.

Levi-Armstrong:
> Does this have anything to do with the Jacobian issue, or is the purpose of this to save on computation in the cases where there is no tolerance applied?

No, just computation and address the issue @marip8 pointed out.

trajopt/src/problem_description.cpp
marrts: Makes sense, and I'll remove that, but this appears to be not used at all anyway.

Levi-Armstrong: Yea, but for this to work we will need to implement or update these existing ones to correctly compute the jacobian.

marrts:
> I would update this so the class holds a std::function which should be used to calculate the error. Then based on construction if toleranced we point it to one function and if not tolerance it points to another function. The function signature should be the same as `calcTransformError` -> `VectorXd err = tesseract_common::calcTransformError(target_tf, source_tf);`

I added an error function option, but I'm not sure what we want to put the default values of the tolerances at now. The error function defaults to `tesseract_common::calcTransformError` if none is given.

trajopt/src/problem_description.cpp
marrts: Removed all the tolerances from the Jacobians, but haven't made it so the Jacobians are actually used yet.

marrts:
> but I'm not sure what we want to put the default values of the tolerances at now.

Just realized I don't actually use the tolerances anywhere any more. Should the tolerances just be completely dropped from these structs/classes and put the burden on a higher level function calling them to pass in an error function that captures the tolerances to calculate the error?

trajopt/src/problem_description.cpp
marrts: You seem to say the Jacobian issue is resolved in #139, but I don't see how/where this was resolved.

marrts: Maybe it was fixed in #266 but no one ever changed `problem_description.cpp` to reflect this.

Levi-Armstrong:
Sorry for the confusion. I would not pass the error function into the constructor. I would keep the tolerance as input to the constructor and have the constructor evaluate the tolerance to determine which function to assign to the internal member variable.

trajopt/src/kinematic_terms.cpp
marrts: That comparison you referenced is doing something slightly different. I reimplemented this though in the latest commit.

trajopt/src/problem_description.cpp
marrts: In my testing my plans always fail if I pass dfdx into the cost and/or constraint. The changes to `kinematic_terms.cpp` in #266 do not seem to have resolved the issue. @Levi-Armstrong, any ideas on what's wrong [here](https://github.com/tesseract-robotics/trajopt/blob/806a40a7b48ac21bbee935f1f8eac3974d2385c6/trajopt/src/kinematic_terms.cpp#L158)?

Levi-Armstrong: We cannot use the existing one because there is a known issue which needs to be resolved. We need to do the same thing that the current jacobian from error function does but use the error function which does not take tolerances into account.

marrts:
This has been working for me now and I think it addresses everything. It has been very useful in increasing motion planning success.

trajopt/src/kinematic_terms.cpp
```diff
@@ -175,6 +175,7 @@ CartPoseErrCalculator::CartPoseErrCalculator(tesseract_kinematics::JointGroup::C
   , indices_(std::move(indices))
 {
   assert(indices_.size() <= 6);
+  is_target_active_ = manip_->isActiveLinkName(target_frame_);
```
marrts, line 178: It appears that `CartPoseErrCalculator` was previously never actually checking if the target was active. This was why the tests were failing for me. The error calculations for the Jacobian and the regular error were in the opposite direction of each other. Not sure if this was potentially leading to other issues.

trajopt/src/kinematic_terms.cpp
```diff
@@ -157,48 +239,33 @@ void CartPoseErrCalculator::Plot(const tesseract_visualization::Visualization::P
 
 MatrixXd CartPoseJacCalculator::operator()(const VectorXd& dof_vals) const
```
Levi-Armstrong, line 240: This should be updated to look similar to TrajOpt Ifopt `CartPosConstraint::CalcJacobianBlock` the issue below is that calcTransformError should not be used for jacobian calculation because it normalizes the rotational error [-pi, pi]. See `tesseract_common::calcRotationalError2`

Levi-Armstrong (changes requested):

trajopt/src/kinematic_terms.cpp
Levi-Armstrong: It's probably a good idea to add a new function to tesseract_common like `calcTransformErrorJac' and update the documentation for `calcTransformError` to have a doxygen warning stating it should not be used for numerical jacobians.

Levi-Armstrong: I created this [PR](https://github.com/tesseract-robotics/tesseract/pull/971) adding the new function.

trajopt/src/kinematic_terms.cpp
```diff
@@ -66,56 +133,71 @@ void DynamicCartPoseErrCalculator::Plot(const tesseract_visualization::Visualiza
...
+  , target_frame_(std::move(target_frame))
+  , source_frame_offset_(source_frame_offset)
+  , target_frame_offset_(target_frame_offset)
+  , indices_(std::move(indices))
+{
+  assert(indices_.size() <= 6);
+  is_target_active_ = manip_->isActiveLinkName(target_frame_);
+
+  // Check to see if the waypoint is toleranced and set the error function accordingly
+  if ((lower_tolerance.size() == 0 && upper_tolerance.size() == 0) || lower_tolerance.isApprox(upper_tolerance, 1e-6))
```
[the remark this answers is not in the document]
marrts: I will have it throw an error if the upper and lower tolerances sizes don't match.

trajopt/src/kinematic_terms.cpp
marrts: Updated, will have to re-run the CI after that PR is merged in.

marrts:
The test that if failing is comparing the numerical to the analytical Jacobian solutions, which would be expected to be different now because of the addition of `calcTransformErrorJac`. Fundamentally the "numerical" and "analytical" solution are doing the same thing, except "analytical" now uses `calcTransformErrorJac` and numerical still uses `calcTransformError` as a basis for the calculation. 

Failing comparison: https://github.com/marrts/trajopt_ros/blob/9b7676069257930aca365c23adc142579a6d3e8d/trajopt/test/kinematic_costs_unit.cpp#L56

Levi-Armstrong:
> The test that is failing is comparing the numerical to the analytical Jacobian solutions, which would be expected to be different now because of the addition of `calcTransformErrorJac`. Fundamentally the "numerical" and "analytical" solution are doing the same thing, except "analytical" now uses `calcTransformErrorJac` and numerical still uses `calcTransformError` as a basis for the calculation.
> 
> Failing comparison: https://github.com/marrts/trajopt_ros/blob/9b7676069257930aca365c23adc142579a6d3e8d/trajopt/test/kinematic_costs_unit.cpp#L56

@marrts I will take a look at this and push a fix.

trajopt/src/problem_description.cpp
```diff
@@ -920,25 +926,37 @@ void CartPoseTermInfo::hatch(TrajOptProb& prob)
...
+    auto f = std::make_shared<CartPoseErrCalculator>(prob.GetKin(),
+                                                     source_frame,
+                                                     target_frame,
+                                                     source_frame_offset,
+                                                     target_frame_offset,
+                                                     indices,
+                                                     lower_tolerance,
+                                                     upper_tolerance);
 
     // This is currently not being used. There is an intermittent bug that needs to be tracked down it is not used.
```
rjoomen, line 938: Is this issue solved now? Otherwise please add @todo, so it can be found more easily.

trajopt/src/problem_description.cpp
```diff
@@ -920,25 +926,37 @@ void CartPoseTermInfo::hatch(TrajOptProb& prob)
...
```
rjoomen, line 955: Is this issue solved now? Otherwise please add @todo, so it can be found more easily.

trajopt/src/problem_description.cpp
```diff
@@ -777,8 +777,14 @@ void DynamicCartPoseTermInfo::hatch(TrajOptProb& prob)
...
+    auto f = std::make_shared<DynamicCartPoseErrCalculator>(prob.GetKin(),
+                                                            source_frame,
+                                                            target_frame,
+                                                            source_frame_offset,
+                                                            target_frame_offset,
+                                                            indices,
+                                                            lower_tolerance,
+                                                            upper_tolerance);
 
     // This is currently not being used. There is an intermittent bug that needs to be tracked down it is not used.
```
rjoomen, line 789: Is this issue solved now? Otherwise please add @todo, so it can be found more easily.

rjoomen:
Could you implement this for trajopt_ifopt as well?

marrts:
> > The test that is failing is comparing the numerical to the analytical Jacobian solutions, which would be expected to be different now because of the addition of `calcTransformErrorJac`. Fundamentally the "numerical" and "analytical" solution are doing the same thing, except "analytical" now uses `calcTransformErrorJac` and numerical still uses `calcTransformError` as a basis for the calculation.
> > Failing comparison: https://github.com/marrts/trajopt_ros/blob/9b7676069257930aca365c23adc142579a6d3e8d/trajopt/test/kinematic_costs_unit.cpp#L56
> 
> @marrts I will take a look at this and push a fix.

This test is failing because the rotational error set to `[-pi,pi]` for the numerical and `[0,2pi]` for the analytical. The test produces these error vectors for the initial error.
```
err_numerical:  -1.19071 -0.0193624  -0.417557  -0.910973  -0.810396   -2.52978
err_analytical:  -1.19071 -0.0193624  -0.417557    1.12722    1.00277    3.13031
```
The Jacobian results are:
```
Error:   Numerical:
  0.0193634  -0.189402   0.199863  0.0145331  0.0711069 -0.0978927          0
 -0.189711    0.37213   0.287726 -0.0535873   0.130371   0.122383          0
         0   0.168799 -0.0644466   -0.33596   0.101585 -0.0885404          0
  0.628394  -0.259181  -0.148265    0.46934   -1.30136   0.384817   0.316577
 -0.256931    1.32699   0.362308   -1.27522  -0.361558   -0.60032   -1.19339
  0.856022   0.134472   -1.03897  -0.181689   0.104203    -0.8853   0.628396
Error:   Analytical:
  0.0193634  -0.189402   0.199863  0.0145331  0.0711069 -0.0978927          0
 -0.189711    0.37213   0.287726 -0.0535873   0.130371   0.122383          0
         0   0.168799 -0.0644466   -0.33596   0.101585 -0.0885404          0
 -0.123759   0.625531  -0.454852   -0.85613    1.29629   -1.09011  -0.156293
  0.899549   -1.37083   -1.01615    1.33296   0.168067   0.196658    1.68613
  0.756401   0.680107  -0.486995  -0.539909   -1.00088  -0.609487  -0.123762
```
The first three rows, corresponding to position, match and the last three, corresponding to rotation are different. This is the expected behavior based on the initial error vectors we have.

If you rotate the target frame by 180 degrees around z this goes away and the test passes:
```
Eigen::Isometry3d target_frame_offset = Eigen::Isometry3d::Identity() * Eigen::AngleAxisd(M_PI, Eigen::Vector3d::UnitZ());
```
```
err_numerical:  -1.19071 -0.0193624  -0.417557  -0.591328   0.664716   0.344705
err_analytical:  -1.19071 -0.0193624  -0.417557  -0.591328   0.664716   0.344705
```

I see a few options for resolving this:
1. Simply apply my change and the test passes
2. Do two tests, one with my change and one without and on the one without just check the first three rows of the matrix match
3. Remove this test altogether because it actually isn't necessarily expected that they match
4. Change the numerical implementation to somehow switch to an error that always has a positive angle

trajopt/src/problem_description.cpp
marrts: I would say this resolves it, so I'm removing those comments.

marrts:
> Could you implement this for trajopt_ifopt as well?

I haven't used trajopt_ifopt yet, and it's not obvious to me what changes would need to be made. Perhaps we get this one merged we make an issue to implement this in trajopt_ifopt?

I do see that there is a `bounds_` object in the `CartPosConstraint` class, does this allow tolerance by default?
https://github.com/tesseract-robotics/trajopt/blob/07337759d7c04cf837c9d2a4fef199c47ed84a74/trajopt_ifopt/include/trajopt_ifopt/constraints/cartesian_position_constraint.h#L175-L176

Levi-Armstrong:
@marrts I would just apply your change to make the unit test pass.

Levi-Armstrong (approved):

Issue #353, opened 2023-11-01
cant understand how to scale the penalty coefficient in ”trajopt_optimizers/trajopt_sqp“.


sailuluo:
in ”trajopt_optimizers\trajopt_sqp\src\trust_region_sqp_solver.cpp“，
It seems that it is necessary to update the constraint coefficient at the end of each loop through "TrustRegionSQPSolver::adjustPenalty()",
But I can only find updates of the SQP result，cant find updates of qp_problem.

Levi-Armstrong:
This is a know issue, I believe this is one the final remaining items to tackle to where the Ifopt version aligns with the legacy. I have not had time to look into how best solve this. If you have any thoughts please let me know.

Levi-Armstrong:
Addressed by #366

PR #366, opened 2023-11-26
Fix TrajOpt Ifopt Numerical IK unit test

Levi-Armstrong:


rjoomen:
You found the cause?

I'll test this tomorrow, but before merging all debug modes should be switched off again.

Levi-Armstrong:
I think so. The only difference with compare mode enabled was the difference in what was used for Infinity. The old version used `std::numeric_limits<double>::infinity()` where the trajopt ifopt used `OSQP_INFTY` which is `1e30` so testing the difference now to see if that makes a difference.

Levi-Armstrong:
I have found another bug with how the constraint merit coefficient is handled in the new version. The best merit is not being updated when the merit coefficient is increase which causes it continue to fail because it compares the previous best to the new best and the new best will never better because than previous best because it is not updated to account for the new constraint merit coeffs. I will have to do some work with comparing this to the previous version to fix correctly.

rjoomen:
Great finds!
Does this also fully solve #353?

rjoomen:
The numerical IK tests now succeed for me on Jammy/Humble.
(The std::scoped_lock build error is likely a C++ build version issue, as the correct header <mutex> is included. This feature has been introduced in C++17.)

Levi-Armstrong:
This does solve issue #353

Levi-Armstrong:
> The numerical IK tests now succeed for me on Jammy/Humble. (The std::scoped_lock build error is likely a C++ build version issue, as the correct header is included. This feature has been introduced in C++17.)

Yea, I tried increasing the CXX version to 17 but it still fails and not sure why. I will merge this and open another issue for this. Would you have time to take a look since you are running Jammy locally?

rjoomen:
> > The numerical IK tests now succeed for me on Jammy/Humble. (The std::scoped_lock build error is likely a C++ build version issue, as the correct header is included. This feature has been introduced in C++17.)
> 
> Yea, I tried increasing the CXX version to 17 but it still fails and not sure why. I will merge this and open another issue for this. Would you have time to take a look since you are running Jammy locally?

I'll look into it later this week.

PR #369, opened 2023-11-28
Handle solver failures like TrajOpt does

### LINKED ISSUE #368: TrajOptIfopt lacks max_qp_solver_failures parameter and logic
TrajOptIfopt has no logic to handle repeated solver failures (with [parameter `max_qp_solver_failures` like TrajOpt](https://github.com/tesseract-robotics/trajopt/blob/master/trajopt_sco/include/trajopt_sco/optimizers.hpp#L105)).

I have a PR in the works to add this.

rjoomen:
- Added max_qp_solver_failures parameter and the logic of TrajOpt to handle solver failures
- Added updateBounds() call after each box size modification
- Moved setVariables() call after a solver failure to solveQPProblem()
- Renamed SQPStatus::TIME_LIMIT to SQPStatus::OPT_TIME_LIMIT because of a name collision with OSQP (in tesseract_planning)

Closes #368

### DIFF
--- a/trajopt_optimizers/trajopt_sqp/src/trust_region_sqp_solver.cpp
+++ b/trajopt_optimizers/trajopt_sqp/src/trust_region_sqp_solver.cpp
@@ -98,7 +98,7 @@ void TrustRegionSQPSolver::solve(const QPProblem::Ptr& qp_problem)
     results_.convexify_iteration = 0;
 
     // Convexification loop
-    for (int convex_iteration = 0; convex_iteration < 100; convex_iteration++)
+    for (int convex_iteration = 1; convex_iteration < 100; convex_iteration++)
     {
       double elapsed_time = std::chrono::duration<double, std::milli>(Clock::now() - start_time).count() / 1000.0;
       if (elapsed_time > params.max_time)
@@ -412,7 +412,7 @@ void TrustRegionSQPSolver::printStepInfo() const
   std::printf("\n| %s |\n", std::string(88, '=').c_str());
   std::printf("| %s %s %s |\n", std::string(36, ' ').c_str(), "ROS Industrial", std::string(36, ' ').c_str());
   std::printf(
-      "| %s %s %s |\n", std::string(32, ' ').c_str(), "TrajOpt Ifopt Motion Planning", std::string(31, ' ').c_str());
+      "| %s %s %s |\n", std::string(28, ' ').c_str(), "TrajOpt Ifopt Motion Planning", std::string(29, ' ').c_str());
   std::printf("| %s |\n", std::string(88, '=').c_str());
   std::printf("| %s %s (Box Size: %-3.9f) %s |\n",
               std::string(26, ' ').c_str(),
--- a/trajopt_optimizers/trajopt_sqp/src/trust_region_sqp_solver.cpp
+++ b/trajopt_optimizers/trajopt_sqp/src/trust_region_sqp_solver.cpp
@@ -383,6 +383,8 @@ SQPStatus TrustRegionSQPSolver::solveQPProblem()
   }
   else
   {
+    qp_problem->setVariables(results_.best_var_vals.data());
+
     CONSOLE_BRIDGE_logError("Solver Failure");
     return SQPStatus::QP_SOLVER_ERROR;
   }
--- a/trajopt_optimizers/trajopt_sqp/src/trust_region_sqp_solver.cpp
+++ b/trajopt_optimizers/trajopt_sqp/src/trust_region_sqp_solver.cpp
@@ -230,6 +230,7 @@ bool TrustRegionSQPSolver::stepSQPSolver()
 void TrustRegionSQPSolver::runTrustRegionLoop()
 {
   results_.trust_region_iteration = 0;
+  int qp_solver_failures = 0;
   while (results_.box_size.maxCoeff() >= params.min_trust_box_size)
   {
     if (SUPER_DEBUG_MODE)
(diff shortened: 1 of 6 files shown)


rjoomen:
I'm not entirely sure about removing the setVariables, what do you think, @Levi-Armstrong? 

The calls to updateBounds() seems necessary to me, as box size changes do modify the bounds [here](https://github.com/tesseract-robotics/trajopt/blob/master/trajopt_optimizers/trajopt_sqp/src/trajopt_qp_problem.cpp#L768).

Levi-Armstrong:
> I'm not entirely sure about removing the setVariables, what do you think, @Levi-Armstrong?
> 
> The calls to updateBounds() seems necessary to me, as box size changes do modify the bounds [here](https://github.com/tesseract-robotics/trajopt/blob/master/trajopt_optimizers/trajopt_sqp/src/trajopt_qp_problem.cpp#L768).

Yea, the set variables need to stay because it need to set them back to the previous iterations values.

trajopt_optimizers/trajopt_sqp/src/trust_region_sqp_solver.cpp
```diff
@@ -240,12 +241,34 @@ void TrustRegionSQPSolver::runTrustRegionLoop()
 
     // Solve the current QP problem
     status_ = solveQPProblem();
+
     if (status_ != SQPStatus::RUNNING)
     {
-      qp_problem->setVariables(results_.best_var_vals.data());
```
Levi-Armstrong, line 245: This just needs to be added back.

rjoomen: Done

rjoomen:
I'm just not sure about what clang-format complains about, as the check is not verbose. On my PC it seems OK.

trajopt_optimizers/trajopt_sqp/src/trust_region_sqp_solver.cpp
Levi-Armstrong: I don't think that would work since it would need to be set before convexify is called in stepSQPSolver. I think we might be able to move into that method but I would want to do a lot of testing to make sure everything aligns with the legacy version before making a change like this.

rjoomen: Isn't this equivalent? In the previous version solveQPProblem() is called, and if qp_solver->solve() failed and hence solveQPProblem() failed setVariables is called from withing runTrustRegionLoop(). Now I moved this line to solveQPProblem(), still to be called if qp_solver->solve() failed. It's just clearer now, as solveQPProblem() either sets the variables to new_var_vals and back to best_var_vals if succeeded and immediately to best_var_vals if failed. See line 386 in my PR.

rjoomen:
Unstable/jammy fails due to clang-tidy-14. It needs a more extensive clang-tidy config, I think. Ubuntu/jammy succeeds, as no clang-tidy is found.

trajopt_optimizers/trajopt_sqp/src/trust_region_sqp_solver.cpp
Levi-Armstrong: Okay, It looks like we already handle setting back the values in solveQPProblem as you pointed out so it should not be needed. I would just recommend running a one of the test which is in both trajopt and trajopt ifopt that does not have a collision cost/constraint to verify that the output tables are the same. I mostly look at the bottom sum values for each iteration.

Levi-Armstrong:
> Unstable/jammy fails due to clang-tidy-14. It needs a more extensive clang-tidy config, I think. Ubuntu/jammy succeeds, as no clang-tidy is found.

Lets try for CI to install the same version of clang-tidy that is used on focal for now until we have time to go through and upgrade to the latest version of clang-tidy.

trajopt_optimizers/trajopt_sqp/src/trust_region_sqp_solver.cpp
rjoomen: Sure, can you point me to which ones?

Levi-Armstrong: I would start with the numerical IK and maybe the simple collision unit test because it is not using the weighted collision approach.

rjoomen: I'm still at this. There are differences, but not caused by this PR. I'm trying to find the root cause.

Levi-Armstrong: I know the numerical ik has difference on Jammy because on one interaction it returns iteration limit instead of approximate solution but focal produces identical results. I am currently running focal so let me know if you want me to compare since I think you are running jammy.

rjoomen: Aha, I'm running in that issue exactly! But I've also access to focal (I'm using WSL), so I'll run it there tomorrow.

Levi-Armstrong: From what I could tell both hit the iteration limit but because of slight numerical differences focal picks approximate solution and jammy pick iteration limit.

Levi-Armstrong: I will take a look

Levi-Armstrong:
I compared trajopt_sqp numerical ik output between this branch and master with no differences found. I think your changes are good.

Levi-Armstrong (approved):

Levi-Armstrong:
@rjoomen You good with me squash merging this?

rjoomen:
Go ahead! Thanks for checking

PR #371, opened 2023-11-30
Set clang tidy version to 10 for CI build

Levi-Armstrong:


rjoomen:
If you add this to .clang-tidy, the scoped_lock errors will be gone:
```
ExtraArgs:
  - '-std=c++17'
```

PR #391, opened 2024-03-04
Remove unneeded reserve() calls

rjoomen:
In expr_ops.hpp reserve() was called at every variable update. This causes the variable and coefficient vectors to be relocated very often, as their size is updated by a minimal amount every time . Removing the reserve() calls allows the vectors to use their normal growing behavior, improving performance.

Before:
image

After:
image


rjoomen:
I'm not sure. Do you know what size the coeff and var vectors typically are? The normal automatic vector reserve growing behavior is by a fixed factor of 1.5, so in the worst case we only use 1.5 times the amount of memory compared to the current situation.

Levi-Armstrong (approved):

rjoomen:
I just read that a shrink_to_fit can also cause a reallocation. So unless we really need the memory it's not advisable to use it.

PR #396, opened 2024-04-05
No longer rely on this repos tags to match tesseract tags for docker

marrts:
Per discussion [here](https://github.com/tesseract-robotics/tesseract_ros2/pull/97#discussion_r1553837824)

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -52,12 +52,24 @@ jobs:
             type=raw,event=pr,prefix=${{ matrix.distro }}-,value=master
             type=semver,pattern={{major}}.{{minor}},prefix=${{ matrix.distro }}-
 
+      # Set TAG environment variable to specify right version of tesseract docker image to pull
+      - name: Set TAG environment variable
+        run: |
+          # Check if the current workflow run was triggered by a tag.
+          if [[ "${{ github.ref_type }}" == "tag" ]]; then
+            # If it's a tag event, set the TAG to the latest tagged version of tesseract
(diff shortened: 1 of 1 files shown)


marrts:
The plan will be to continue this forward for each subsequent tesseract repo. So tesseract planning would have the new step like this:
```yaml
 - name: Set TAG environment variable
        run: |
          # Check if the current workflow run was triggered by a tag.
          if [[ "${{ github.ref_type }}" == "tag" ]]; then
            # If it's a tag event, set the TAG to the latest tagged version of trajopt
            echo "TAG=${{ matrix.distro }}-0.7" >> $GITHUB_ENV
          else
            # If it's not a tag event (e.g., a push to a branch or a PR), set the TAG to distro-master.
            echo "TAG=${{ matrix.distro }}-master" >> $GITHUB_ENV
          fi
```

And then this would eventually address [this](https://github.com/tesseract-robotics/tesseract_qt/issues/103) issue because we can specify master and don't have to worry about primary branch names matching across repos.

Levi-Armstrong:
Since tesseract is a dependent I would prefer going forward that this have the same tag as the rest of the tesseract repos for the same reason we decided for those repos.

marrts:
> Since tesseract is a dependent I would prefer going forward that this have the same tag as the rest of the tesseract repos for the same reason we decided for those repos.

Should we also change tesseract_qt main to be master to keep consistency then? And keep main but mark it as deprecated? Then we could simplify this docker tag issue to just match tags all the way through.

Levi-Armstrong:
That is fine with me.

Levi-Armstrong:
I renamed tesseract_qt branch

PR #398, opened 2024-04-10
Fix mismatched delete/free in OSQPModel

Jack-ReframeSystems:
Previously we were calling delete on the CSC matrix pointer but its malloc'd, we should be calling "free" instead. We also don't want to free any of the subfield (using something like `csc_spfree`) since we manage the individual fields in OSQPModel vectors.

### DIFF
--- a/.run-clang-format
+++ b/.run-clang-format
@@ -1,2 +1,2 @@
 #!/bin/bash
-find . -type f -regex '.*\.\(cpp\|hpp\|cc\|cxx\|h\|hxx\)' -exec clang-format-8 -style=file -i {} \;
+find . -type f -regex '.*\.\(cpp\|hpp\|cc\|cxx\|h\|hxx\)' -exec clang-format -style=file -i {} \;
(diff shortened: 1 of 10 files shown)


Levi-Armstrong (changes requested):
LGTM, Could can you run clang format? There is a script .run-clang-format that you may run.

Jack-ReframeSystems:
> LGTM, Could can you run clang format? There is a script .run-clang-format that you may run.

Do you mind if I change the clang format script to use a newer version of clang? Its using clang 8 at the moment but I can't get that on Ubuntu Jammy, I would like to change it to use the clang default version on the system. 

I'll update the PR with this and you can let me know if you have concerns with this.

.run-clang-format
```diff
@@ -1,2 +1,2 @@
 #!/bin/bash
-find . -type f -regex '.*\.\(cpp\|hpp\|cc\|cxx\|h\|hxx\)' -exec clang-format-8 -style=file -i {} \;
```
Levi-Armstrong, line 2: Just for consistency please revert change.

Levi-Armstrong:
replaced by #399

rjoomen:
Good catch!

PR #401, opened 2024-04-29
Copy IFOPT code locally to allow customization and add nodes variable set

Levi-Armstrong:


rjoomen:
Wouldn't a custom `tesseract-robotics` Ifopt fork be better? Adding it all to Trajopt inflates the code base, whereas having it in a separate project also helps with code separation etc.

PR #411, opened 2024-07-12
Trajopt clang-tidy clean-up

rjoomen:


Levi-Armstrong (changes requested):
The only thing I request is to use std::size_t instead of size_t.

rjoomen:
Done

Levi-Armstrong (approved):

PR #417, opened 2024-07-27
Add boost-stacktrace vcpkg package to Windows Github Action

johnwason:


Levi-Armstrong:
Windows error `fatal error C1083: Cannot open include file: 'boost/format.hpp': No such file or directory`

PR #431, opened 2024-11-30
Add trajopt_ifopt builds for Windows and MacOS

johnwason:
Thanks to @traversaro the `coin-or-ipopt` package is now available on vcpkg, allowing for `trajopt_ifopt` to build on Windows and MacOS. The builds are currently failing. I think the problem on Windows is someone is importing Windows.h which includes a lot of conflicting macros. The conflicting macros include `min()` and `max()`. We can try using `NOMINMAX` to work around this, but I think there are a few other name conflicts causing problems.


johnwason:
Superseded by #559

PR #446, opened 2025-01-22
Update OSQP workspace, instead of recreating

rjoomen:
This recreates the OSQP workspace, instead of recreating it, if the sparsity of the P and A matrices has not changed.

Below are runtime results of some examples (average of three runs each). Warm start is switched on or off using `trajopt_solver_profile->settings.warm_start`.

| | puzzle_piece_example | puzzle_piece_auxillary_axes_example |
| -- | -- | -- |
| Cleanup/setup | 10.09463 | 28.40169 |
| Update, warm start | 7.164561 | 24.4617 |
| Update, cold start | 7.944992 | 24.38523 |

@Levi-Armstrong can you take a good look at this? The examples do show increased performance, but we'll have to be sure there are no side effects. In my own work I did not see any improvements, but my optimization times are much lower.


Levi-Armstrong:
It might be good to make this configurable to allow time for testing. Is that something that can easily be done?

rjoomen:
> It might be good to make this configurable to allow time for testing. Is that something that can easily be done?

I just made this possible, this also requires https://github.com/tesseract-robotics/tesseract_planning/pull/596

Levi-Armstrong:
Should these changes also be added to trajopt_ifopt?

Levi-Armstrong:
I fine if we want to merge this now and add an issue to port to Ifopt.

rjoomen:
Do you want to merge this, or do you want to do some testing yourself? For my use case this does actually reduce performance somewhat (both in time and success rate), so it would be great if you can validate this using your more extensive knowledge of Trajopt.
You can also merge and then do the testing, as the default [here](https://github.com/tesseract-robotics/tesseract_planning/pull/596) has this disabled, so it should not impact anyone who does not explicitly enable this.

rjoomen:
> Should these changes also be added to trajopt_ifopt?

They should, once we have validated them for Trajopt. For TrajoptIfopt the implementation will be somewhat different, as we use osqp-eigen to interface with OSQP.

trajopt_sco/src/osqp_interface.cpp
```diff
@@ -122,6 +124,16 @@ void OSQPModel::updateObjective()
...
+  // Check if sparsity has changed
+  bool sparsity_equal = false;
+  if (osqp_data_.P != nullptr && osqp_data_.P->n == osqp_data_.n && osqp_data_.P->m == osqp_data_.n &&
+      osqp_data_.P->nzmax == P_csc_data_.size())
+  {
+    sparsity_equal = memcmp(osqp_data_.P->p, P_column_pointers_.data(), static_cast<size_t>(osqp_data_.P->n) + 1) == 0;
+    sparsity_equal = sparsity_equal &&
+                     (memcmp(osqp_data_.P->i, P_row_indices_.data(), static_cast<size_t>(osqp_data_.P->nzmax)) == 0);
+  }
+
```
Levi-Armstrong, line 136: Should we check sparsity_equal if not requested?

rjoomen: Done

Levi-Armstrong:
@rjoomen With this disabled you did not see any performance impact? I wanted to check before merging.

Levi-Armstrong:
> > @rjoomen With this disabled you did not see any performance impact? I wanted to check before merging.
> 
> I'll run some tests again tomorrow to make sure and get back to you.
> 
> Have you been able to run your own tests already, or are you waiting for the merge? I was really hoping this would give a performance boost, but that didn't happen.

Yea, if all this does is avoid a few malloc calls, then it may be in the noise.

rjoomen:
It should enable warm starting the optimization, leading to faster convergence.

Levi-Armstrong:
Do you do any checking with how often the sparsity does not change? I would assume with how collision cost/constraint works that it is always changing. It my be better to run a test using the inverse kinematics test because the sparsity should never change.

rjoomen:
Now updating only happens when the previous optimization was solved. I also added the option to explicitly warm start using the previous solution when not updating. Both updating and explicitly warm starting gives me worse results than always recreating the workspace.

BTW, I just now cleaned this code up a bit more, but have not been able to test those changes, as I'm not up to date with all Tesseract changes of the last three weeks yet. Once the TrajOptCollisionConfig modifications have stabilized in all Tesseract project I'll update my forks and my own projects and retest this PR.

rjoomen:
> Did you do any checking with how often the sparsity does not change? I would assume with how collision cost/constraint works that it is always changing. It my be better to run a test using the inverse kinematics test because the sparsity should never change.

It is quite often that the sparsity does not change. But if in the IK tests the sparsity never changes, this might be a better application for update/warm start than typical path planning problems.

Levi-Armstrong:
@rjoomen How has the latest version been working out?

rjoomen:
Haven't tried yet, as I was waiting for https://github.com/tesseract-robotics/tesseract_ros2/pull/149

Levi-Armstrong (approved):

PR #457, opened 2025-04-07
Removed weighted sum methods from legacy Trajopt

marip8:
This PR removes the weighted sum collision methods from legacy `trajopt`. A number of issues were addressed with these methods during the implementation of `trajopt_ifopt` but they were never backported to legacy `trajopt`. Since the legacy `trajopt` will ultimately be sunset in favor of `trajopt_ifopt`, it doesn't make sense to go through the effort of backporting the fixes. Instead, we will remove them from legacy `trajopt` so users don't run into issues trying to use these methods that are known to have bugs.


Levi-Armstrong:
> Wasn't this supposed to be fixed by replacing it by the weighted sum methods of Trajopt-Ifopt, instead of being dropped completely? This would also allow for better comparisons between legacy Trajopt and Trajopt-Ifopt.

The plan is still to replace this with how it is currently done in TrajOpt Ifopt. We decided to just remove the code from legacy for now because it is not correct and should not be used.

PR #469, opened 2025-08-08
Add yaml conversions and related tests

marrts:
For CollisionCoeffData and TrajOptCollisionConfig


Levi-Armstrong (changes requested):
I think it would be best to update CollisionCoeffData to resemble the CollisionMarginData

PR #471, opened 2025-08-19
Fix benchmarks

rjoomen:
This fixes the benchmarks, which were outdated.

This also revives the use of the [TRAJOPT_ENABLE_RUN_BENCHMARKING ](https://github.com/tesseract-robotics/trajopt/blob/b7909c57afd22fc2fdf7c50e43df4d297e834fba/README.md?plain=1#L97) cmake variable, to allow building the benchmarks without having to wait for them to be ran immediately.

This also adds building the benchmarks (without running them) to the CI, to prevent them becoming outdated again.


rjoomen:
The update of .clang-tidy disables two checks that are also disabled in Tesseract. The rest of the modifications is just reordering to match Tesseract when comparing side by side.

PR #472, opened 2025-08-21
Fix bounds problem

rjoomen:
This fixes a problem with incorrect bounds. 

Example:
```
lb = -2.5
ub = 2.5
trust_box_size = 0.0015
x = 2.6
```

Then `lbtrust = 2.5985`, while `ubtrust = 2.5`, causing an error in OSQP, as the lower bound is larger than the upper bound.

With this PR, the resulting `lbtrust = 2.4985`, which is smaller than the upper bound.

### DIFF
--- a/trajopt_sco/src/optimizers.cpp
+++ b/trajopt_sco/src/optimizers.cpp
@@ -95,13 +95,13 @@ void BasicTrustRegionSQP::setTrustBoxConstraints(const DblVec& x)
   const VarVector& vars = prob_->getVars();
   assert(vars.size() == x.size());
   const DblVec& lb = prob_->getLowerBounds();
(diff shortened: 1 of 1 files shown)


rjoomen:
The second commit of this PR lowers the solve time for the `tesseract_examples_puzzle_piece_auxillary_axes_example` for me from >24s to >18s with Bullet (both simple and BVH). With FCL I see a comparable improvement.

Levi-Armstrong:
It looks like the same issue might exist in the IFOPT version; can you take a look and make the same change if needed? It looks like this logic happens in `void TrajOptQPProblem::Implementation::updateNLPVariableBounds()`

Levi-Armstrong (approved):

PR #553, opened 2026-04-07
Reduce allocation overhead and fix thread safety in collision evaluators

rjoomen:
(Disclosure: Created with help of Claude.)

  - Clone JointGroup per CollisionEvaluator (759ebbc6) — Each evaluator now owns a cloned JointGroup instead of sharing one. This eliminates mutex contention in calculateTransforms under OMP concurrency, and fixes a latent data race where calcJacobianHelper calls JntToJac — which mutates KDL's internal Joint::posecaching state — without holding the mutex. Perf: lll_lock_wait 0.29→0.13%, KDL::Joint::pose 0.70→0.31%.
  - Move ContactResultMap instead of deep copying (53544b2a) — Replace the thread_local + clear() + deep copy pattern in GetContactResultCached with a per-call local moved into make_shared. The previous approach copied every Rb_tree node, string key, and ContactResult vector on every cache miss. Perf:_Rb_tree::_M_copy (0.46%) and _M_erase (0.43%) eliminated entirely, memmove -0.35%, allocator group -0.29%. Total ~1.5%.
  - Use persistent FK transform caches (41bb694f) — Convert the static thread_local TransformMap caches to instance members and stop clearing them before each FK call. operator[] overwrites existing entries in-place, so clearing only adds deallocation and reallocation overhead. Adds transforms_diff_update_ to separate diff-block and manip-block entries, avoiding the need to clear between phases. Perf: TransformMap::clear() (0.90%) and _M_insert_unique_node (0.14%) eliminated, but operator[] increased from 1.56→2.25% (+0.69%) because hash lookups on pre-populated maps are more expensive than inserts into empty ones. Net ~0.5%.
  - Add rvalue overloads for exprInc/exprDec (4d03f60f) — exprInc(AffExpr&, const AffExpr&) copies the vars vector, bumping atomic refcounts on every shared_ptr<VarRep>. The new rvalue overloads use make_move_iterator to transfer elements instead. Also adds std::move at call sites in CalcDistExpressions*Free where local vectors were passed as lvalues despite being discarded immediately after. Perf: _M_range_insert -0.12%, _Sp_counted_base::_M_release -0.09%. Total ~0.25%.

  Related: tesseract state_solver fix

  The persistent FK caches above rely on the state solver overwriting existing TransformMap entries rather than skipping them. The KDL solver already does this via operator[], but the OFKT solver used insert() which silently skips existing keys. This was fixed separately in tesseract (https://github.com/tesseract-robotics/tesseract/pull/1271), with a corresponding removal of redundant TransformMap::clear() calls in kinematic_terms.cpp (b87a1711). Perf: TransformMap::clear() 1.38→0.99% (-0.39%), but operator[] 1.72→1.97% (+0.25%) for the same reason as above. Net ~0.17%.

  Profiling

  Measured on a private planning setup (cast collision, large kinematic tree, OMP concurrency) using perf. Each optimization was applied incrementally and profiled in isolation against the preceding baseline.

  The net picture is that eliminating allocation-heavy patterns (clear/rebuild, deep copy) shifts cost toward hash lookups on pre-populated maps. The savings outweigh the added lookup cost, but the margin is smaller than the headline numbers suggest: e.g. the FK cache optimization eliminates 1.04% but adds back 0.69% in operator[], for a net of ~0.5%.

  Overall: ~5% wall-clock reduction, with 2.4% of CPU symbols eliminated entirely, offset by ~0.7% increased operator[] cost and measurement noise in other symbols. The JointGroup clone (mutex + correctness) and ContactResultMap move (deep copy elimination) are the largest clear wins.

  Profiling

  Measured on a private planning setup (cast collision, large kinematic tree, OMP concurrency) using perf. Each optimization was applied incrementally and profiled in isolation against the preceding baseline. Overall: ~5% wall-clock reduction, with 2.4% of CPU symbols eliminated entirely, 1.7% significantly reduced, and 0.4% less allocator overhead.

### DIFF
--- a/trajopt/include/trajopt/collision_terms.hpp
+++ b/trajopt/include/trajopt/collision_terms.hpp
@@ -238,10 +238,11 @@ protected:
 
   std::pair<ContactResultMapConstPtr, ContactResultVectorConstPtr> GetContactResultCached(const DblVec& x);
 
-  static thread_local tesseract::common::TransformMap transforms_cache0;         // NOLINT
-  static thread_local tesseract::common::TransformMap transforms_cache1;         // NOLINT
-  static thread_local tesseract::common::TransformMap transforms_cache0_update;  // NOLINT
-  static thread_local tesseract::common::TransformMap transforms_cache1_update;  // NOLINT
+  tesseract::common::TransformMap transforms_cache0_;
+  tesseract::common::TransformMap transforms_cache1_;
+  tesseract::common::TransformMap transforms_diff_update_;
+  tesseract::common::TransformMap transforms_cache0_update_;
+  tesseract::common::TransformMap transforms_cache1_update_;
 
   void CollisionsToDistanceExpressions(sco::AffExprVector& exprs,
                                        std::vector<double>& exprs_margin,
--- a/trajopt/src/collision_terms.cpp
+++ b/trajopt/src/collision_terms.cpp
@@ -481,7 +481,7 @@ void CollisionEvaluator::CalcDistExpressionsStartFree(const DblVec& x,
   for (std::size_t i = 0; i < exprs0.size(); ++i)
   {
     exprs[i] = sco::AffExpr(dist_results[i].get().distance);
-    sco::exprInc(exprs[i], exprs0[i]);
+    sco::exprInc(exprs[i], std::move(exprs0[i]));
     exprs[i] = sco::cleanupAff(exprs[i]);
   }
 }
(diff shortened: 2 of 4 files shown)


rjoomen:
@Levi-Armstrong, what might interest you is that I got a significant performance increase using [jemalloc](https://github.com/jemalloc/jemalloc). I know you have used tcmalloc, but that somehow caused crashes for me.

Levi-Armstrong (approved):

PR #554, opened 2026-04-13
Move from string-based links and joints to id-based logic (using hashes)

rjoomen:
Requires https://github.com/tesseract-robotics/tesseract/pull/1274

For reference: this was squash merged, the full 61 commit history is backed up here:
https://github.com/rjoomen/trajopt/commits/archive/pr-554-pre-squash


Levi-Armstrong:
@rjoomen I have merged in the Tesseract PR; once the remaining PR's have been rebased I will do a review and get these merged in.

rjoomen:
Update: All CI is green (except the stable CI, as expected: not a problem), https://github.com/tesseract-robotics/tesseract/pull/1339 fixed the last alignment problems in the CI.

trajopt/include/trajopt/collision_terms.hpp
```diff
@@ -84,7 +84,7 @@ struct CollisionEvaluator
   using ConstPtr = std::shared_ptr<const CollisionEvaluator>;
 
   // NOLINTNEXTLINE
-  CollisionEvaluator(std::shared_ptr<const tesseract::kinematics::JointGroup> manip,
+  CollisionEvaluator(const std::shared_ptr<const tesseract::kinematics::JointGroup>& manip,
```
Levi-Armstrong, line 87: Good find; I see that we were using this after a move.

trajopt/src/kinematic_terms.cpp
```diff
@@ -378,19 +378,20 @@ MatrixXd CartVelJacCalculator::operator()(const VectorXd& dof_vals) const
...
+  MatrixXd jac0;
+  MatrixXd jac1;
   jac0.resize(6, manip_->numJoints());
   jac1.resize(6, manip_->numJoints());
 
   if (tcp_.translation().isZero())
   {
-    jac0 = manip_->calcJacobian(dof_vals.topRows(n_dof), manip_->getBaseLinkName(), link_);
-    jac1 = manip_->calcJacobian(dof_vals.bottomRows(n_dof), manip_->getBaseLinkName(), link_);
+    jac0 = manip_->calcJacobian(dof_vals.topRows(n_dof), link_id_);
```
Levi-Armstrong, line 388: check: This appears to be using a different overload because the base_link is not being provided

trajopt_ifopt/include/trajopt_ifopt/constraints/collision/continuous_collision_evaluators.h
```diff
@@ -78,20 +78,6 @@ class ContinuousCollisionEvaluator
...
-   * @param dof_vals1 Joint values for end state
-   * @param contact_results The contact results for transition between dof_vals0 and dof_vals1 to compute the gradient
-   * data
-   * @return The gradient results
-   */
-  virtual trajopt_common::GradientResults
-  calcGradientData(const Eigen::Ref<const Eigen::VectorXd>& dof_vals0,
-                   const Eigen::Ref<const Eigen::VectorXd>& dof_vals1,
-                   const tesseract::collision::ContactResult& contact_results) = 0;
-
```
Levi-Armstrong, line 94: check: Why were these removed?

trajopt_ifopt/include/trajopt_ifopt/constraints/collision/discrete_collision_evaluators.h
```diff
@@ -95,16 +95,6 @@ class DiscreteCollisionEvaluator
...
-
-  /**
-   * @brief Extracts the gradient information based on the contact results
-   * @param dofvals The joint values
-   * @param contact_result The contact results to compute the gradient
-   * @param isTimestep1 Indicates if this is the second timestep when computing gradient for continuous collision
-   * @return The gradient results
-   */
-  virtual trajopt_common::GradientResults getGradient(const Eigen::VectorXd& dofvals,
-                                                      const tesseract::collision::ContactResult& contact_result) = 0;
```
Levi-Armstrong, line 107: check: Why was this removed?

Levi-Armstrong (commented):
@rjoomen I have done a first pass with a few questions. There are a few files I want to go through in more detail but wanted to get these questions out to speed up the process.

trajopt/src/kinematic_terms.cpp
rjoomen: Intentional, and equivalent. The 2-arg overload is documented as returning the jacobian *relative to the joint group base link*, which is exactly what the old `calcJacobian(vals, getBaseLinkName(), link_)` asked for — and the 3-arg overload short-circuits to it: `if (base_link_id == getBaseLinkId()) return calcJacobian(joint_angles, link_id);`. The only thing dropped is a `LinkId -> string -> LinkId` round-trip per call in this hot `operator()`.

trajopt_ifopt/include/trajopt_ifopt/constraints/collision/continuous_collision_evaluators.h
rjoomen: Their only real work was a *per-contact* margin lookup that belongs *per-pair*. The caller already iterates the pair-keyed `ContactResultMap` with the ordered `LinkIdPair` in scope and already does `coeff_data_.getCollisionCoeff(pair.first)` there, so the margin lookup moved next to it and the free `trajopt_common::getGradient` is now called directly. Equivalence is exact: `getCollisionMargin(id1, id2)` just builds the ordered key and forwards to `getCollisionMargin(pair)`, and `pair.first` *is* that key. That left the virtuals as trivial pass-throughs, so they went.

They were public API on the abstract evaluators; there are no callers in tesseract_planning, tesseract_ros2 or tesseract_qt, but I'm happy to restore them as non-virtual helpers if you'd rather not break the interface.

trajopt_ifopt/include/trajopt_ifopt/constraints/collision/discrete_collision_evaluators.h
rjoomen: Same reason as the continuous one: it was a trivial wrapper whose only work was a per-contact margin lookup, now hoisted to once per link pair in the caller (which already has the ordered `LinkIdPair` key in scope for the coeff lookup) with the free `trajopt_common::getGradient` called directly. Same offer applies: it can come back as a non-virtual helper if you want the interface preserved.

trajopt_common/src/collision_types.cpp
```diff
@@ -38,37 +38,79 @@ void CollisionCoeffData::setDefaultCollisionCoeff(double default_collision_coeff
...
+
+  lookup_table_.insert_or_assign(pair, collision_coeff);
 
   if (tesseract::common::almostEqualRelativeAndAbs(collision_coeff, 0.0))
-    zero_coeff_.insert(key);
+    zero_coeff_.insert(pair);
   else
-    zero_coeff_.erase(key);
+    zero_coeff_.erase(pair);
 }
```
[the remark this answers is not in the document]
rjoomen: `LinkIdPair` is `OrderedIdPair`, which canonicalizes in its constructor and in `assign()` -- an unordered pair cannot be constructed, so no ordering is lost. The hash-collision check was indeed unable to observe anything and has been removed.

trajopt_common/include/trajopt_common/yaml_extensions.h
```diff
@@ -52,16 +52,16 @@ struct convert<trajopt_common::CollisionCoeffData>
...
+        key_node.push_back(key.second.name());
 
-        // tell yaml-cpp “emit this sequence in [a, b] inline style”
+        // tell yaml-cpp "emit this sequence in [a, b] inline style"
         key_node.SetStyle(YAML::EmitterStyle::Flow);
 
-        pair_coeff_data_node[key_node] = pair.second;
+        pair_coeff_data_node[key_node] = coeff;
       }
       node["pair_coeff_data"] = pair_coeff_data_node;
```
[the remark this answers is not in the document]
rjoomen: Fixed: the node is no longer hand-built here at all, it goes through tesseract's `LinkIdPairMapConvert`, which constructs a `NodeType::Map`.

trajopt/src/file_write_callback.cpp
```diff
@@ -66,7 +66,7 @@ void WriteFile(const std::shared_ptr<std::ofstream>& file,
       rot_vec(3) = q.z();
 
       // Write cartesian pose to file
-      *file << it.first << ": ";
+      *file << it.first.value() << ": ";
```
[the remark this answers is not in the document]
rjoomen: Fixed: this now writes `it.first.name()`.

trajopt_common/data/contact_manager_plugins.yaml
```diff
@@ -4,8 +4,9 @@ contact_manager_plugins:
   search_libraries:
     - tesseract_collision_bullet_factories
     - tesseract_collision_fcl_factories
+    - tesseract_collision_coal_coal_factories
   discrete_plugins:
-    default: BulletDiscreteBVHManager
+    default: CoalDiscreteBVHManager
```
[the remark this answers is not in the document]
rjoomen: Reverted: Bullet is the default again and the Coal entries are gone from this file.

trajopt_common/data/contact_manager_plugins.yaml
```diff
@@ -21,8 +22,12 @@ contact_manager_plugins:
           max_collision_algorithm_pool_size: 100
       FCLDiscreteBVHManager:
         class: FCLDiscreteBVHManagerFactory
+      CoalDiscreteBVHManager:
+        class: CoalDiscreteBVHManagerFactory
+        config:
+          gjk_guess_threshold: 0.005
   continuous_plugins:
-    default: BulletCastBVHManager
+    default: CoalCastBVHManager
```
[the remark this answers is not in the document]
rjoomen: Reverted: Bullet is the default again and the Coal entries are gone from this file.

trajopt/test/cast_cost_octomap_arm_unit.cpp
```diff
@@ -0,0 +1,385 @@
...
+    return { sco::INVALID, 0.0, -1, 0.0 };
+
+  const std::vector<std::string>& joint_names = prob->GetKin()->getJointNames();
+  const std::vector<std::string>& active_links = prob->GetKin()->getActiveLinkNames();
+  const tesseract::scene_graph::StateSolver::UPtr state_solver = prob->GetEnv()->getStateSolver();
+
+  auto opt = std::make_shared<sco::BasicTrustRegionSQP>(prob);
+  // Few iterations keep runtime manageable (~20s/iter with octree collision checking).
+  // Enough for the optimizer to show avoidance progress if the backend provides gradients.
+  opt->getParameters().max_iter = 3;
```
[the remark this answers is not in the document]
rjoomen: Removed: that test asserted trivially (both backends hit `OPT_FAILED`), and dropping it took the trajopt suite from ~140s to ~42s.

trajopt/src/kinematic_terms.cpp
Levi-Armstrong: This change is fine

trajopt_ifopt/include/trajopt_ifopt/constraints/collision/continuous_collision_evaluators.h
Levi-Armstrong: I am good with the change.

trajopt/src/collision_terms.cpp
```diff
@@ -1305,23 +1286,21 @@ sco::ConvexObjective::Ptr CollisionCost::convex(const sco::DblVec& x, sco::Model
 
 double CollisionCost::value(const sco::DblVec& x)
 {
-  DblVec dists;
-  m_calc->CalcDists(x, dists);
```
Levi-Armstrong, line 1309: Is this okay to remove?

trajopt/src/collision_terms.cpp
```diff
@@ -1390,23 +1366,21 @@ sco::ConvexConstraints::Ptr CollisionConstraint::convex(const sco::DblVec& x, sc
 
 DblVec CollisionConstraint::value(const sco::DblVec& x)
 {
-  DblVec dists;
-  m_calc->CalcDists(x, dists);
```
Levi-Armstrong, line 1394: Same here?

trajopt/src/collision_terms.cpp
rjoomen: Yes. On master `CalcDists` was just

```cpp
const ContactResultVectorConstPtr dist_results = GetContactResultVectorCached(x);
CollisionsToDistances(*dist_results, dists);
```

and `CollisionsToDistances` only pushes `dist_result.get().distance` for each entry, in order. The next line of `value()` already calls `GetContactResultVectorCached(x)` itself, so the old code did the cache lookup twice and built a parallel `DblVec` whose element *i* is by construction `dist_results[i].get().distance` -- exactly what the loop now reads straight off the `ContactResult` it already holds. Same values in the same order, one fewer cache lookup and one fewer heap-allocated vector per `value()` call.

`CalcDists` and `CollisionsToDistances` have no remaining callers, so both are gone.

trajopt/src/collision_terms.cpp
rjoomen: Same, and for the same reason -- see the reply on `CollisionCost::value`: https://github.com/tesseract-robotics/trajopt/pull/554#discussion_r3806741413

trajopt_ifopt/src/constraints/collision/continuous_collision_evaluators.cpp
```diff
@@ -56,36 +56,34 @@ LVSContinuousCollisionEvaluator::LVSContinuousCollisionEvaluator(
...
     };
-    env_active_link_names_ = env_->getActiveLinkNames();
-
-    std::sort(manip_active_link_names_.begin(), manip_active_link_names_.end());
-    std::sort(env_active_link_names_.begin(), env_active_link_names_.end());
-    std::set_difference(env_active_link_names_.begin(),
-                        env_active_link_names_.end(),
-                        manip_active_link_names_.begin(),
-                        manip_active_link_names_.end(),
-                        std::inserter(diff_active_link_names_, diff_active_link_names_.begin()));
```
Levi-Armstrong, line 76: Is the sorting not required?

trajopt_ifopt/src/constraints/collision/continuous_collision_evaluators.cpp
```diff
@@ -291,36 +276,34 @@ LVSDiscreteCollisionEvaluator::LVSDiscreteCollisionEvaluator(
...
```
Levi-Armstrong, line 311: Same here?

trajopt_ifopt/src/constraints/collision/discrete_collision_constraint.cpp
```diff
@@ -198,14 +199,25 @@ int DiscreteCollisionConstraintD::update()
   coeffs_.resize(rows_);
   values_.resize(rows_);
 
-  Eigen::Index i{ 0 };
+  // Iterate in deterministic key order so the QP constraint row layout is
+  // independent of unordered_map bucket ordering (which varies with insertion
+  // order, i.e. the broadphase traversal order of the collision backend).
+  sorted_keys_.clear();
+  sorted_keys_.reserve(collision_data_->contact_results_map.size());
```
Levi-Armstrong, line 206: I think this is why the sorting previously commented on is not needed?

Levi-Armstrong:
@rjoomen I have finished my review with two additional questions.

trajopt_ifopt/src/constraints/collision/continuous_collision_evaluators.cpp
rjoomen: Not required. Those two `std::sort` calls were never about the members themselves -- they were the precondition of `std::set_difference`, which needs both input ranges sorted. All three members were `std::vector<std::string>`; they are now `std::unordered_set<LinkId>`, so the same difference is computed by hash lookup:

```cpp
for (const auto& id : env_active_link_ids_)
  if (manip_active_link_ids_.find(id) == manip_active_link_ids_.end())
    diff_active_link_ids_.insert(id);
```

No ordered-range requirement left, so nothing to sort. Every remaining use of the three members is either a membership test or order-independent iteration: `setCollisionObjectsTransform(link_id, ...)` one call per link, each writing its own object; and `ContactResultMap::addInterpolatedCollisionResults`, whose parameter is `const std::unordered_set<LinkId>&` and which only calls `find()` on it.

(As a sanity check on that: the old `std::inserter(diff_active_link_names_, begin())` on a *vector* produced the difference in reverse sorted order, so nothing downstream can have depended on the ordering either.)

trajopt_ifopt/src/constraints/collision/discrete_collision_constraint.cpp
rjoomen: Different thing, as it happens -- the two are unrelated.

The constructor sorts were a precondition of `std::set_difference` over the active-link vectors; they disappear because those members became `std::unordered_set<LinkId>` and the difference is now a hash-lookup loop.

`sorted_keys_` is about the QP constraint **row layout** when iterating `contact_results_map`, which is an `unordered_map`. `update()` and `getJacobian()` traverse it independently, so if the traversal order shifts between them the values and the jacobian rows desynchronise. `ContactResultMap` became an `unordered_map` in the tesseract identity migration (`5e8ad73bb`), so before this the row layout followed bucket order and could reshuffle between SQP iterations, misaligning OSQP's row-indexed warm-start duals. This mirrors what `ContinuousCollisionConstraintD` already did.

trajopt_ifopt/src/constraints/collision/continuous_collision_evaluators.cpp
rjoomen: Same, and for the same reason -- see the reply on `LVSContinuousCollisionEvaluator`: https://github.com/tesseract-robotics/trajopt/pull/554#discussion_r3807667687

rjoomen:
> @rjoomen I have finished my review with two additional questions.

All replied to!

PR #564, opened 2026-05-29
Updated debian package CI build

marip8:
Updates the debian package build CI job to be triggered correctly for past versions. Corresponds to https://github.com/tesseract-robotics/tesseract_planning/pull/750

### DIFF
--- a/.github/workflows/package_debian.yml
+++ b/.github/workflows/package_debian.yml
@@ -7,45 +7,30 @@ on:
 
   # allow manually starting this workflow
   workflow_dispatch:
+    inputs:
+      major_version:
+        description: 'Major Version'
+        required: true
+        type: string
+      minor_version:
+        description: 'Minor Version'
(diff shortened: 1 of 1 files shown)


Levi-Armstrong:
@marip8 Does this change now require it to only be manually triggered or does it still work as expected when generating a release?

marip8:
It probably won't work when a new tag is pushed to the repo because it extracts the required version information from user inputs which probably only get populated by a user dispatch.

That said, I don't think there is a currently a purpose to running this job when a new tag is pushed because the action does not upload the artifacts anywhere, and Github deletes them after a certain amount of time. I don't think I'm opposed to making this a fully manual CI job because it seems like the debians are currently being attached to releases manually anyway.

The other option is to make this job run on releases only (not pushed tags), and to find a way to upload the debs as an artifact to the release for which it was run. That seems like it should be possible

PR #574, opened 2026-08-10
Fix: getClosestFeasiblePoint returns a point outside narrow bounds

rjoomen:
Follow-up to #573.

The `delta` inset is applied unconditionally, so when a variable's bounds are narrower than `2 * delta` the result lands *outside* them — a "closest feasible point" that is infeasible. For zero-width bounds (`lb == ub`, e.g. a locked joint) the result is `ub - delta`, i.e. below the lower bound.

Capping the inset at half the bound width places the point at the midpoint instead, which is as far inside as the bounds allow:

```cpp
const double inset = std::min(delta, (upper_bounds_[i] - lower_bounds_[i]) / 2);
y[i] = std::min(std::max(x[i], lower_bounds_[i] + inset), upper_bounds_[i] - inset);
```

Behaviour is unchanged for any bounds wider than `2 * delta`. Writing it as a single expression also removes the two-assignment pattern that caused #573. `std::clamp` would read better, but it has a `lo <= hi` precondition that inverted bounds would turn into UB, whereas this form is total.

`modeling.cpp` had no direct unit test, so this adds `modeling-unit.cpp` covering both bounds, the inset, narrow and zero-width bounds, infinite bounds, and idempotence. Against current master only the narrow-bounds test fails; the other four pass and serve as regression guards for #573.

### DIFF
--- a/trajopt_sco/include/trajopt_sco/modeling.hpp
+++ b/trajopt_sco/include/trajopt_sco/modeling.hpp
@@ -235,9 +235,9 @@ public:
   DblVec getCentralFeasiblePoint(const DblVec& x);
   /**
    * @brief Pushes the input vector within joint limits by some amount in a repeatable fashion
-   * @param x Input vector. If it is already within joint limits, it will not be changed
+   * @param x Input vector. Values further than delta inside the joint limits are not changed
    * @param delta (Default = 1e-3) Amount that it is pushed within the joint limits. Pushing it exactly onto a limit can
-   * cause numerical problems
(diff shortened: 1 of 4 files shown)


Levi-Armstrong (approved):
This can be merged after the conflicts are fixed and CI is passing.

rjoomen:
> This can be merged after the conflicts are fixed and CI is passing.

Needs https://github.com/tesseract-robotics/trajopt/pull/554 first, as the tesseract container for CI is now id-based. Will rebase and fix once that one has landed.

PR #579, opened 2026-08-21
Restore clear() on the thread-local FK transform caches

rjoomen:
Follow-up to tesseract-robotics/tesseract_planning#734, where @Levi-Armstrong pointed out that a
`static thread_local` forward-kinematics scratch map that is never cleared holds the **union of every
link set the thread has served**. A planning server running many environments over one thread pool
grows it without bound. The same problem exists here.

`trajopt/src/kinematic_terms.cpp` lost five `clear()` calls in 43b3cbad (mine, merged before #554's
squash, so it is not part of 39e7cb00). Its reasoning covered correctness only: `calcFwdKin` does
overwrite every key it owns, which says nothing about keys left behind by a previous environment.
`transforms_cache` stayed `static thread_local` on `TrajOptVectorOfVector` / `TrajOptMatrixOfVector`,
so the residency is unbounded.

Four commits, each independently droppable:

1. **Restore the five clears.** One per call, not per FK evaluation — the finite-difference loops
   reuse the populated map deliberately, exactly as they did before 43b3cbad.
2. **Read the caches with `at()` instead of `operator[]`** (14 sites). `operator[]`
   default-constructs an `Isometry3d`, whose matrix is *uninitialised*, and multiplies that into the
   result with no diagnostic. Nothing validates the frames at construction — the constructors check
   only `isActiveLinkId`, which is false both for a static link and for a link absent from the group.
   The keys are present in practice because `JointGroup::calcFwdKin` assigns every scene link on
   every call, but that is precisely the assumption #734 was right to push on; `at()` makes the code
   fail loudly instead of returning a garbage pose. This is a behaviour change on malformed input,
   hence its own commit.
3. **Regression test.** `KinematicCostsTest.TransformsCacheDoesNotAccumulateAcrossEnvironments`
   evaluates a `CartPoseJacCalculator` against the 89-link `arm_around_table` group and then the
   7-link `boxbot` one on the same thread, and asserts each call sees only its own group's links.
   Two *scene graphs* are needed — two groups over one graph write identical key sets, so the union
   never grows. Verified to fail when the clear is removed again: the second call then sees 95 links
   rather than 7.

   It observes the cache through `error_diff_function_`, which `operator()` hands the map it has
   just populated. That is deliberate — see the Windows note below.
4. **Document the contract** on the two declarations, which carried no comment at all.

### On the alternative

Moving the cache into the five calculators as a `mutable` member — the shape used in 184926b4 for
`CollisionEvaluator`, and in `trajopt_ifopt/costs/squared_cost.h` — would keep the allocation saving
and bound residency. I rejected it: `static thread_local` stays correct if a calculator instance is
ever shared across threads, and a `mutable` member does not, with nothing enforcing the invariant.
A per-instance per-thread cache keeps both properties but needs `omp_get_thread_num()`, and `trajopt`
core does not link OpenMP; its footprint is also wrong by roughly two orders of magnitude. Happy to
revisit if you prefer the member form.

To be precise about what the fix buys: `clear()` deallocates the nodes but keeps the bucket array, so
node storage returns to the current link set while the bucket count stays at its per-thread
high-water mark. That is 8 B/bucket against ~152 B/node, so the residual is the small part — but the
map does not return to nothing.

### One Windows observation

The first version of this test derived from `TrajOptVectorOfVector` to read `transforms_cache`
directly. That links on Linux through default visibility, but `conda-win` rejected it:

```
error LNK2019: unresolved external symbol
  "protected: static ... trajopt::TrajOptVectorOfVector::transforms_cache"
```

Neither base class nor the member carries `TRAJOPT_API`, so the symbol is not exported from
`trajopt.dll`. All sixteen classes deriving from those two bases live inside the library, so nothing
is broken today — but the member is `protected` in a public header, which reads as an invitation to
derive, and any out-of-DLL deriver would hit the same wall. I did not add the export: widening the
Windows ABI to make a test link seemed the wrong trade in this PR. Instead the test observes the map
through `error_diff_function_`, which needs no exported symbol. Say the word if you would rather
have the export.

### Out of scope, flagged only

- **Constructor validation.** `JointGroup::hasLinkId` exists and is unused here; validating the
  frames up front would give a better message than a bare `std::out_of_range`. It is a design call
  across five classes, so not in this PR.
- **`trajopt_ifopt` has a redundant `calcFwdKin`.** `CartPosConstraint::calcJacobianBlock` (:332) and
  `CartLineConstraint::calcJacobianBlock` (:195) call `calcFwdKin` a second time with the identical
  `joint_vals`, recomputing a cache the top of the same function already populated; in the line
  constraint four locals then shadow the outer ones with identical values. Not a residency problem —
  both functions do clear — just wasted work. Untouched here.
- **`trajopt_ifopt` reads its caches with `operator[]`** at 7 sites, the same hazard commit 2 removes
  from `kinematic_terms.cpp`. Left out to keep commit 2 droppable; happy to add it here or in a
  separate PR.

Tests: `trajopt`, `trajopt_sco`, `trajopt_ifopt`, `trajopt_sqp`, `trajopt_common` all green (189
tests, 0 failures).

### DIFF
--- a/trajopt/src/kinematic_terms.cpp
+++ b/trajopt/src/kinematic_terms.cpp
@@ -99,8 +99,8 @@ VectorXd DynamicCartPoseErrCalculator::operator()(const VectorXd& dof_vals) cons
 {
   transforms_cache.clear();
   manip_->calcFwdKin(transforms_cache, dof_vals);
-  const Isometry3d source_tf = transforms_cache[source_frame_] * source_frame_offset_;
-  const Isometry3d target_tf = transforms_cache[target_frame_] * target_frame_offset_;
+  const Isometry3d source_tf = transforms_cache.at(source_frame_) * source_frame_offset_;
+  const Isometry3d target_tf = transforms_cache.at(target_frame_) * target_frame_offset_;
 
   VectorXd err = error_function(target_tf, source_tf);
 
@@ -161,8 +161,8 @@ MatrixXd DynamicCartPoseJacCalculator::operator()(const VectorXd& dof_vals) cons
   // Finite-difference jacobian of the (optionally toleranced) Cartesian error.
   transforms_cache.clear();
   manip_->calcFwdKin(transforms_cache, dof_vals);
-  const Isometry3d source_tf = transforms_cache[source_frame_] * source_frame_offset_;
-  const Isometry3d target_tf = transforms_cache[target_frame_] * target_frame_offset_;
+  const Isometry3d source_tf = transforms_cache.at(source_frame_) * source_frame_offset_;
+  const Isometry3d target_tf = transforms_cache.at(target_frame_) * target_frame_offset_;
 
   Eigen::MatrixXd jac0(indices_.size(), dof_vals.size());
   Eigen::VectorXd dof_vals_pert = dof_vals;
@@ -170,8 +170,8 @@ MatrixXd DynamicCartPoseJacCalculator::operator()(const VectorXd& dof_vals) cons
   {
     dof_vals_pert(i) = dof_vals(i) + epsilon_;
     manip_->calcFwdKin(transforms_cache, dof_vals_pert);
-    const Isometry3d source_tf_pert = transforms_cache[source_frame_] * source_frame_offset_;
-    const Isometry3d target_tf_pert = transforms_cache[target_frame_] * target_frame_offset_;
+    const Isometry3d source_tf_pert = transforms_cache.at(source_frame_) * source_frame_offset_;
+    const Isometry3d target_tf_pert = transforms_cache.at(target_frame_) * target_frame_offset_;
     VectorXd error_diff = tesseract::common::calcJacobianTransformErrorDiff(
         target_tf, target_tf_pert, source_tf, source_tf_pert, lower_tolerance_, upper_tolerance_);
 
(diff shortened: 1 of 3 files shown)


trajopt/src/kinematic_terms.cpp
```diff
@@ -413,11 +417,13 @@ VectorXd CartVelErrCalculator::operator()(const VectorXd& dof_vals) const
...
 
+  transforms_cache.clear();
+
   manip_->calcFwdKin(transforms_cache, dof_vals.topRows(n_dof));
-  Isometry3d pose0 = transforms_cache[link_id_] * tcp_;
+  Isometry3d pose0 = transforms_cache.at(link_id_) * tcp_;
 
   manip_->calcFwdKin(transforms_cache, dof_vals.bottomRows(n_dof));
-  Isometry3d pose1 = transforms_cache[link_id_] * tcp_;
+  Isometry3d pose1 = transforms_cache.at(link_id_) * tcp_;
```
Levi-Armstrong, line 426: Do you think there would be any significant performance difference using `.at` given this can be called a lot?

rjoomen: No - I measured it rather than guess. On the 89-link `arm_around_table`/`right_arm` map (RelWithDebInfo, best of 7, and with the two arms run in both orders to rule out ordering bias):

| | per lookup |
|---|---|
| `operator[]` | 3.33 - 3.49 ns |
| `.at()` | 3.52 - 3.53 ns |

About 0.15 ns apart, which is roughly one branch: on a hit both do a single find, and the difference is `at()`'s throw path against `operator[]`'s insert path. The heaviest call site here does ~16 lookups, so that is a couple of nanoseconds against an `operator()` costing microseconds - around 0.04%. The safety is effectively free.

The same probe did turn up a number worth passing on, since it is about this PR rather than about `.at`. Restoring the clear costs three orders of magnitude more than the lookups do: `calcFwdKin` into an already-populated map is 2.6 us, `clear()` + `calcFwdKin` is 5.9 us, so the clear adds ~3.3 us per call - 89 node allocations through `Eigen::aligned_allocator`. That is exactly what 43b3cbad was buying, now with a number on it.

Correctness and bounded residency are worth that, and I would keep this PR as it stands. It does suggest the scratch map wants a better design eventually - bounded without paying for a fresh allocation every call - but that is a separate conversation, not this PR.

Levi-Armstrong: LGTM, Thanks for running the test

Levi-Armstrong (approved):

PR #586, opened 2026-09-05
Batch the collision evaluator transform updates

rjoomen:
Caller half of tesseract-robotics/tesseract#1361, which adds bulk collision object operations to the
contact manager base classes. **Merge that one first** — these call sites do not compile without the
new overloads.

### What changes

Two starting points, one destination.

The sco evaluators in `trajopt/src/collision_terms.cpp` already issued a single bulk call — they
copied the active links' poses into a scratch update map and passed that map. What goes away there
is the hand-rolled copy: they now pass the active link ids and the FK cache directly.

The trajopt_ifopt evaluators did loop the flushing single-object setter, one call per link per
state. Those now issue one bulk call, so a backend that batches performs one broadphase flush per
state instead of one per link.

### Performance

Around +3.7% on Coal for these `ALL` + distance + penetration queries, against a measured ~10 us
saved per continuous contact check. As with the other caller PRs this is a stack-level figure — the
batching itself lands in tesseract#1361. Bullet is unaffected.

### Note for subclass authors

Three `protected` members are removed from `CollisionEvaluator`:
`transforms_diff_update_`, `transforms_cache0_update_` and `transforms_cache1_update_`. They were
the scratch update maps described above and have no remaining purpose. Nothing in the workspace
referenced them, and they were never public surface, but an out-of-tree subclass touching them would
notice. A new `all_active_link_ids_` member replaces the repeated union of the manipulator and diff
active link ids.

### Testing

Full workspace across the five affected repositories: 1908 tests, 0 failures, 14 pre-existing skips
(3 of them in `trajopt_sco`, pre-existing and unrelated).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


Levi-Armstrong (approved):

rjoomen:
https://github.com/tesseract-robotics/tesseract_planning/issues/768

PR #588, opened 2026-09-09
Evaluate the collision gradient at the states and witness point the check used

rjoomen:
Neither stack's collision gradient is the derivative of the distance the collision check reports.
Three things differ: the pose the reference offset is rotated by, the states the jacobians are taken
at, and the point the offset points to. This PR fixes all three, in both stacks, so that each
computes Eq. (20) of Schulman et al. 2014 per cast, referenced to the witness point the check
actually found.

Measured against central differences of the checker's own distance, the old gradient carries 17 %
median error at a cast of 0.8 rad, and 43 % at a contact pinned to a cast end. The corrected one is
at finite-difference noise except for one documented gap, which needs a change upstream of this
repository. The discrete evaluator, which most examples ship with, is byte-for-byte unchanged.

## The defect

**Rotation.** `jacobianChangeRefPoint` needs the offset to the witness point in the world frame,
so the local point is rotated by the link's pose. That pose must belong to the configuration the
jacobian was evaluated at. All three sites instead used a pose the `ContactResult` carries, and a
subdivided check stores the poses of the sub-segment the contact was found in, not of the state
being linearised about. Both stacks had this error.

**Linearisation point.** A contact found by a cast lies on the hull the link sweeps between the
cast's two end states, and the distance the checker reports, which sco's merit uses, is the
distance to that hull. Its derivative, Eq. (20), takes the start half's jacobian at the cast's
start state and the end half's at its end state, weighted by where the contact lies in the cast.
sco did that on an unsubdivided segment, but under subdivision it took the jacobians at the segment
endpoints instead of at the ends of the cast the contact was found in. ifopt took one jacobian at
the interpolated `cc_time` state for both halves: the derivative of the distance along the
joint-space path, not of the hull the checker measures. On single casts, relative error of each
form against central differences of the checker's own distance, 400 samples per row:

| cast length | Eq. (20), median / p90 | one jacobian at `cc_time`, median / p90 | Eq. (20) closer |
|---|---|---|---|
| 0.05 | 0.09 % / 0.28 % | 0.10 % / 0.42 % | 72 % |
| 0.2 | 0.40 % / 1.19 % | 1.59 % / 11.7 % | 85 % |
| 0.4 | 0.85 % / 2.20 % | 8.21 % / 28.3 % | 94 % |
| 0.8 | 1.38 % / 3.55 % | 16.7 % / 42.5 % | 99 % |

**Reference point.** The offset must also point at the right point, and `ContactResult` names the
witness twice: `nearest_points[i]` is the witness in world coordinates, `nearest_points_local[i]` a
point in the frame of `transform[i]`, the pose the contact carries. Both backends guarantee only
that `transform[i]` maps one to the other — not that the local point is the witness's own
coordinates on the link. For a contact typed `CCType_Time0` or `CCType_Time1`, `transform[i]` is
the *start* of the cast whichever end the contact is pinned to, so the stored local point is the
world witness taken back through a pose it does not lie in. Rotating it into the pinned end's pose
then drags the reference point along the whole sweep: 2.8 % of the gradient at a cast of 0.05 rad,
rising to 43 % median and 88 % p90 at 0.8. All three sites had this error, and it is independent of
the two above — it reaches any cast check, subdivided or not.

Where the contact's own time names the state the jacobian is evaluated at, the check found the
contact with the link there, so the witness lies on the link at that state and `nearest_points[i]`
locates it exactly. Everywhere else the witness lies inside the volume the link sweeps, on neither
end's surface, and the stored local point stands in for it as a fixed point of the link.

sco's first two errors partly cancelled. With one cast per segment the sub-segment poses are the
endpoint poses, so its rotation error vanishes; under subdivision its gradient was off by 5.4 %
mean and 52 % max against a numerical reference on `arm_around_table`. Fixing sco's rotation alone
therefore made some scenes worse.

After this PR both stacks use the same model: each end of a contact's cast contributes the
jacobian at its own state, rotated by that state's pose, referenced to the witness point that
state carries, and weighted by where the contact lies in the cast.

## The gradient

Write a segment as $q(t) = q_0 + t\,(q_1 - q_0)$, and take a contact on link $i$ with normal $n$,
sign $\sigma_i$ ($-1$ on the first link), and `cc_time` $c$. Eq. (20) linearises the hull distance
of one cast at the cast's two end states, weighted by where in the cast the contact lies:

$$\frac{\partial d}{\partial q_0} = (1-c)\,\sigma n^{\mathsf T} J(q_0), \qquad \frac{\partial d}{\partial q_1} = c\,\sigma n^{\mathsf T} J(q_1)$$

The paper's $\alpha$ is $1-c$; Bullet computes $c$ from the same ratio of distances to the two
ends' support points. Under subdivision the contact lies in cast $[a,b] = [k/N, (k{+}1)/N]$, where
$N$ is `castCount` and $k = \lfloor cN \rfloor$, at local position $\tau = (c-a)/(b-a)$. Applying
Eq. (20) to that cast and chaining through $q(a) = (1-a)q_0 + a\,q_1$ and $q(b)$:

$$\frac{\partial d}{\partial q_0} = \sigma n^{\mathsf T}\left[(1-\tau)(1-a)\,J(q(a)) + \tau(1-b)\,J(q(b))\right]$$

$$\frac{\partial d}{\partial q_1} = \sigma n^{\mathsf T}\left[(1-\tau)\,a\,J(q(a)) + \tau\,b\,J(q(b))\right]$$

Those four coefficients are `intervalWeights`. Each timestep stores its two weights' sum as `scale`
and their weighted mean as `gradient`; every consumer multiplies the two. Three cases reduce:

- $N = 1$, a `CONTINUOUS` check or a segment no longer than `lvs`: $a = 0$, $b = 1$, $\tau = c$, so
  the pair above is Eq. (20) itself.
- A link typed `CCType_Time0` or `CCType_Time1`: the interval is $[c,c]$ with $c \in \{0,1\}$, so
  one timestep carries the whole gradient at its own endpoint. This is the paper's case where every
  hull vertex comes from one pose and the discrete formula applies, which is also the test Bullet
  uses to assign those types.
- Cast count 0, a check at interpolated states: the interval is $[c,c]$, both halves are taken at
  $q(c)$, and the result is what a discrete check already produced.

### Which point the gradient is referenced to

Each $J$ above is the link jacobian moved to the witness point, so each end needs that point in its
own frame. `intervalWeights` reports whether either end of the cast carries the contact's own time,
which is where the witness is on the link and `nearest_points[i]` gives it exactly; the other end
falls back to the stored local point.

The test is on the contact's **time**, not its `cc_type`. `addInterpolatedCollisionResults`
re-types a sub-cast's `Time1` to `Between` unless the sub-cast is the segment's last, and a
`Between` contact whose `cc_time` lands on a cast boundary has its witness on one end's support
point just as a pinned contact does. On a subdivided PR2 segment, 23 of 81 link-contacts have
their `cc_time` at an end of their cast; a `cc_type` gate catches almost none of them. Their
reported witness sits up to 13 mm from the stored local point read as a fixed point of the link.

Nor is $\tau$ compared for exact equality: tesseract core forms the contact time as
`(sub_index + local) * dt` while the interval ends here are `index / count`, so an end carrying the
contact's time typically misses it by a rounding in either direction.

Three readings are available at such a boundary and they disagree by up to 23 mm, so the choice was
measured — median / p90 relative gradient error against central differences, PR2 segment run
sub-cast by sub-cast, 300 and 302 samples:

| reading | `Between` at `cc_time` 0 | `Between` at `cc_time` 1 |
|---|---|---|
| rotate the stored local point | 0.000 % / 1.209 % | 0.000 % / 0.000 % |
| `transform[i] * nearest_points_local[i]` | 0.000 % / 1.209 % | 3.013 % / 5.167 % |
| `nearest_points[i]` | **0.000 % / 0.000 %** | **0.000 % / 0.000 %** |

`nearest_points[i]` is the only reading exact in all three regimes — these two and the pinned
contact — so that is what the implementation reads. The residual 1.209 % p90 in the first column is
the averaged support point of the next section, which reading the reported witness removes.

### What the model does not capture

- The normal and the contact's position in its cast are held fixed while differentiating, as in the
  paper, which notes the distance is non-differentiable where they switch.
- The swept volume is approximated by the convex hull of the cast's two poses, which *underestimates*
  it when the link rotates, so a cast check can miss a rotational sweep. The paper's remedy is an arc
  padding term; the Coal backend implements one, Bullet does not. Either way the gradient here is the
  derivative of the distance the checker reports.
- Away from the contact's own time the two ends share one witness point, the mean of their two
  support points, because `ContactResult` carries a single `nearest_points_local` per link where the
  paper differentiates each end's own point. This is the whole of Eq. (20)'s residual above: on a
  sphere-sphere cast scene, giving each end its own support point drops the residual from 0.09 %
  median at the shortest cast and 1.38 % at the longest to finite-difference noise at every length,
  while the paper's fixed convex combination costs nothing measurable. Both backends compute the two
  support points and average them, and the mean alone does not determine either, so closing this is a
  change to the contact data upstream, not to this gradient.
- Subdivision is outside the paper, which casts a segment once. Applying Eq. (20) per cast and
  chaining to the waypoints assumes nothing beyond the chain rule, and reduces to the paper on one
  cast.

## What changed

The two stacks share no executed gradient code, so each fix reaches exactly one of them; the
helpers that place a contact in its cast are shared.

**Shared** (`trajopt_common`): `castCount` gives the number of casts a continuous check splits a
segment into, `contactInterval` the cast a contact time lies in, `intervalWeights` how a
link's gradient splits between that cast's two end states and which end, if either, carries the
contact's own time, `intervalState` the configuration at a fraction of the segment, and
`blendIntervalEnds` the weighted mean of the two ends. Each timestep stores the total weight
as `scale` and the weighted mean as `gradient`, so `scale` keeps its meaning for
`GradientResultsSet` and the weighted-average consumers.

**ifopt** (`trajopt_common`, `trajopt_ifopt`): `calcGradient` takes the link transform as a
parameter. The single-state caller passes the stored pose, which is already correct. The two-state
`getGradient` takes the check's cast count and blends the jacobians at the two ends of the
contact's cast, each with its pose computed at that state: two forward-kinematics calls wherever a
contact lies strictly inside its cast. A link typed `CCType_Time0` or `CCType_Time1` is a point in
time at that endpoint. A link the check reported no interval for (negative `cc_time`) is
linearised at the segment start rather than extrapolated outside it; no evaluator produces that
case. The evaluators report the cast count their check uses; the check sub-steps on the same
count.

**sco** (`trajopt`): the three free-endpoint builders (`CalcDistExpressionsStartFree`, `EndFree`,
`BothFree`) build from the two-state gradient, with the same model; the cast and LVS-discrete
evaluators share those entry points. `CalcDistExpressionsSingleTimeStep` keeps the single-state
builder, which rotates by each contact's stored poses as master does, a discrete check leaving
them at the state it checked. The public single-state entry point computes the poses from the
evaluator's state function instead, since a cast contact's stored poses belong to its cast, and
takes its weights from `intervalWeights` over the whole segment so that it splits a cast contact
exactly as the two-state form does. The two-state path computes the poses at each state it
evaluates into a scratch map the evaluator owns, separate from the contact-check caches, which on
a cache hit can hold another state's poses.

**API.** The two-state `trajopt_common::getGradient` gains a trailing `cast_count`.
`trajopt_ifopt::ContinuousCollisionEvaluator` gains the pure virtual `getCastCount`, so a
downstream subclass stops compiling until it reports the cast count its check uses (0 for a check
at interpolated states). `trajopt::CollisionEvaluator` gains the virtual `GetCastCount`, default
0, which `CastCollisionEvaluator` overrides. A `longest_valid_segment_length` of 0 or less counts
as one cast in `castCount`.

## Measurements

Every arm is built against the same underlay, downstream planners and example binaries included:
this PR adds a virtual to both stacks' evaluator base classes, so mixing one arm's libraries under
another's corrupts the heap rather than producing a number.

sco terminates when its trust region collapses, so the **last cost table it prints is a rejected
trial step**. Every cost below is the accepted iterate — that step's `oldexact`, which the solver
also prints as its final `cost values`.

### Car seat

The only shipped example configured `LVS_CONTINUOUS`, so the only one whose collision terms reach
the cast gradient at all. Both stacks, shipped configuration, three interleaved rounds, every round
identical to the last digit.

| stack | master | this branch |
|---|---|---|
| sco | 21 SQP iterations, cost 7.222 | **17 iterations**, cost 7.273 |
| ifopt | 20 iterations, cost 7.118 | **18 iterations**, cost 7.118 |

(Iterations of the car-seat problem; the example solves a trivial freespace problem first, which
takes one iteration on every arm.)

Both stacks converge in fewer iterations. ifopt reaches the same cost; sco's is 0.7 % above
master's. Those are two local minima of a non-convex problem, and this example does not order
iteration count against final cost — an earlier revision of this branch reached the worst cost of
any arm measured, 7.373, in 18 iterations, fewer than master's 21.

### Puzzle piece

437 Cartesian waypoints, 7-DOF iiwa, 3059 variables, sco, single threaded, OSQP at shipped
defaults. Only the collision check type and length vary. Every cell was run twice; the runs agree
on success and on trajectory L1 to every printed digit.

| configuration | master | this branch |
|---|---|---|
| `DISCRETE` (what the example ships) | plans, L1 2211.6900558326 | identical: 21545 lines of verbose log, differing only in the timing line |
| `CONTINUOUS`, one cast per segment | plans, 2211.9788237450 | plans, 2210.5346107376 |
| `LVS_CONTINUOUS`, `lvs` 0.2 | plans, 2210.3765950258 | plans, 2208.5377250498 |
| `LVS_CONTINUOUS`, `lvs` 0.005 | `OPT_PENALTY_ITERATION_LIMIT`, no trajectory | `OPT_PENALTY_ITERATION_LIMIT`, no trajectory |
| `LVS_DISCRETE`, `lvs` 0.005 | plans, 2206.8370214779 | plans, 2209.0968969737 |

The `DISCRETE` row is the check that nothing moved where nothing should.

The `lvs` 0.005 row is where this scene is decided, and it deserves its detail because neither arm
plans it. Both do identical work — 38 function evaluations, 37 QP solves, every cost term zero —
and both end with exactly one of the 437 constraints above the 1.0e-04 tolerance that gates
success:

| arm | worst violation | over tolerance | total violation |
|---|---|---|---|
| master | 7.736e-04 | 1 of 437 | 0.00905 |
| this branch | 1.127e-04 | 1 of 437 | **0.00877** |

So the branch improves the worst waypoint sevenfold and carries the lower total violation, and
still misses a binary gate by 13 % on one waypoint of 437. An earlier revision of this branch
cleared it, at 8.724e-05 — a 13 % margin the other way. That is a threshold crossing, not a
capability difference, and this PR does not claim one: the evidence for the gradient is the tables
above, which compare against the distance the checker itself reports.

`LVS_DISCRETE` at 0.005 failed on master as of 2026-09-10 and plans on today's master; the change
that flipped it was not identified.

### Puzzle piece on the auxiliary axes

The same toolpath with the part on a two-axis positioner, 9 DOF, so both links of a contact can be
active. Trajectory L1, every configuration converging on both arms, two runs agreeing to every
digit. Run on Coal: under Bullet the 0.005 subdivision exhausts 47 GB of memory, where Coal peaks
at 239 MB.

| configuration | master | this branch |
|---|---|---|
| `LVS_CONTINUOUS`, `lvs` 0.005 | 2299.4175096634 | 2299.2954772551 |
| `LVS_CONTINUOUS`, `lvs` 0.2 | 2298.4077575393 | 2298.3973138484 |
| `CONTINUOUS`, one cast per segment | 2298.3455238724 | 2298.4987595622 |
| `DISCRETE` | 2298.4923044785 | identical |

`DISCRETE` is bit-identical, as it must be: no cast, so no interval and no witness to re-read. The
answer moves in every configuration that casts, one-cast included — the reference-point fix reaches
pinned contacts, which a one-cast check produces as readily as a subdivided one. No arm separates
on time, and peak memory is within 2 % of master, the second jacobian per contact.

### arm_around_table (PR2 `right_arm`)

- ifopt / `trajopt_sqp`: **-16.6 %** wall clock (24.7 s -> 20.6 s), median of five interleaved
  rounds, range -17.7 % to -15.4 %, 13 outer iterations in every run.
- sco, 8 one-field perturbations: wall clock range -48 % to +15 % across variants (median +4.5 %);
  `total_cost` agrees within 0.1 % on all 8 and is lower on this branch for 4. The answer is fixed
  while the route to it is not, so this scenario cannot resolve a systematic cost either way.
  Measuring it at all needs OSQP's `max_iter` raised from 8192: at the shipped value both arms
  truncate, and `osqp_interface.cpp` maps `SOLVED_INACCURATE` to `CVX_SOLVED` but
  `MAX_ITER_REACHED` to `CVX_FAILED`, so the arms differ only in which label they draw. Not
  touched here.

## Tests

`trajopt_common/test/collision_gradient_frame_tests.cpp` (10 cases) and
`trajopt/test/cast_gradient_frame_unit.cpp` compare a gradient against a numerical reference: the
witness point's world position, contracted with the normal, differentiated at each cast end. Their
pinned-contact fixtures build the contact the way a backend does — the witness on the link at the
state the check found it, the stored local point that world point taken back through `transform` —
so a reference point rotated out of the stored local point fails them.
`trajopt_common/test/contact_interval_tests.cpp` pins `castCount`, `contactInterval` and
`intervalWeights`, including which end carries the contact's time when the two are a rounding apart
in either direction, and that neither end carries it when the time lies outside the interval.
`trajopt_ifopt/test/continuous_collision_interval_unit.cpp` checks that each ifopt evaluator
reports the cast count its check uses. `cast_expression_frame_unit.cpp` and
`discrete_expression_frame_unit.cpp` check the whole emitted QP expression, per variable block with
endpoint weighting, against a numerical reference sharing no code with the implementation. The cast
expression tests run a subdivided PR2 segment, and a small sphere arm
(`trajopt_common/data/sphere_arm.urdf`) subdivided and on one cast through each of the three
free-endpoint builders, and assert a contact well inside its cast, where a gradient taken at a
single state would differ; the discrete test asserts a `Between` contact. A scene that stopped
producing such a contact fails rather than passes vacuously, and every comparison carries a
non-triviality floor, since `Eigen::isApprox` accepts any pair of near-zero vectors.

The two-state overload's `#ifndef NDEBUG` numerical-jacobian check is removed: every subdivided
contact now goes through it, and it would cost ~2n forward-kinematics evaluations per contact in
CI's Debug builds. The expression-level tests replace it.

## Out of scope

- Giving each cast end its own support point, the last bullet of the model section. It needs
  `ContactResult` to carry both, which is a change upstream of this repository.
- The per-contact forward kinematics runs inside `BasicTrustRegionSQPMultiThreaded`'s parallel
  regions, where every thread shares one cached solver behind a mutex. That optimiser is opt-in
  (`num_threads` defaults to 0); [tesseract#1363](https://github.com/tesseract-robotics/tesseract/pull/1363)
  fixes the serialisation.
- The two LVS-discrete checks still size their sub-stepping from `ceil(dist / lvs) + 1`
  unguarded, so a `longest_valid_segment_length` of 0 or less is undefined behaviour or a count
  of 1 or less there. Pre-existing; the continuous checks now take their count from `castCount`,
  which counts it as one cast.


rjoomen:
@Levi-Armstrong this is ready. Some CI fails because of the plugin loader update, but the relevant ones succeed

Levi-Armstrong:
@rjoomen I want to look at this in more detail to make sure I understand the issue it solving.

rjoomen:
> @rjoomen I want to look at this in more detail to make sure I understand the issue it solving.

@Levi-Armstrong I have updated the PR body to hopefully more clearly explain what was done and why.

rjoomen:
I have an update in the works, as looking at the Schulman Trajopt paper the updated approach of this PR is not entirely correct either. This will land this afternoon.

PR #451, opened 2025-02-23
Leverage tesseract_collision CollisionEvaluatorType

Levi-Armstrong:

PR #590, opened 2026-09-11
Cast CONTINUOUS segments once in the legacy cast evaluator

rjoomen:
Legacy trajopt's `CastCollisionEvaluator` sub-stepped every segment longer than
`longest_valid_segment_length` whatever the configured type, so `CONTINUOUS` behaved as
`LVS_CONTINUOUS`. Gate the sub-stepping on `LVS_CONTINUOUS`, as `trajopt_ifopt`'s
`LVSContinuousCollisionEvaluator`, tesseract's `checkTrajectory` and tesseract_planning's
`contactCheckProgram` already do.

## Behaviour change

A legacy trajopt config set to `CONTINUOUS` now gets one cast per segment. To keep sub-stepping,
set `LVS_CONTINUOUS`. `DISCRETE`, `LVS_DISCRETE` and all of `trajopt_ifopt` are unchanged.

## How it regressed

These semantics were agreed on #451 and implemented in #453, which overrode the length for
`CONTINUOUS` in `CollisionTermInfo::hatch`. #455 moved the length into `TrajOptCollisionConfig`
and stopped passing the override on, so it has been computed and discarded since. Nothing tested
it. This removes the dead computation, and an unused `n_terms` local in `fromJson` that sized the
per-step vectors #455 replaced. The measurement notes in #588 ran into the same defect.

## Tests

Both packages get `ContinuousCastsSegmentOnce`. On a segment about seven configured lengths long,
a `CONTINUOUS` check must return the same contacts as a single cast. The test first asserts that
a sub-stepped check returns more, so the comparison can tell the two apart. The legacy test fails
on master; the `trajopt_ifopt` one pins behaviour that was already correct, and that evaluator's
class doc and sub-stepping comment, which described every segment as interpolated, are corrected.

Suites green: trajopt 44/44, trajopt_ifopt 40/40.


Levi-Armstrong:
LGTM

